PR #316, opened 2015-01-02
Bash/zsh autocompletion functions

o-lim:
Add bash auto-completion functions and update zsh auto-completion functions.


ajacksified:
Thanks! This is awesome. @a-b and I were working on this a while ago, but stopped after we couldn't find a good way to get `luarocks install` to place the file in the right location. Do you have any ideas?

o-lim:
I'm not that familiar with how `luarocks install` works, but I did notice the following at the bottom of the rockspec:
```
 install = {
  bin = {
   ['busted'] = 'bin/busted'
   }
}
```
Is it possible to do something like the following?
```
 install = {
  bin = {
   ['busted'] = 'bin/busted'
  },
  etc= {
   ['bash_completion.d/busted.bash'] = 'completions/bash/busted.bash'
  },
  usr = {
   ['share/zsh/site-functions/_busted'] = 'completions/zsh/_busted'
  },
}
```
Installation for zsh might be a littler trickier, since the installation location is different on different systems (i.e. `/usr/share/zsh/{ver-number}/functions/` on Fedora/CentOS/RedHat vs. `/usr/share/zsh/functions/Completions/busted/` on Ubuntu).  Would have to look at `fpath`, but maybe `/usr/share/zsh/site-functions` is universal?

After doing a quick search, the above may not be possible.  From http://luarocks.org/en/Rockspec_format, it looks like another build back-end may be needed.  Of course, it would only be applied to Unix-like platforms (e.g. `unix`, `macosx`, `mingw32`, but not `win32`).
```
build = {
 type = 'make',
 ...
}
```
or
```
build = {
 type = 'command',
 install_command = 'make install_completion',
}
```
Using the `build.copy_directories` it looks like you may be able to copy into `/usr/*`, but probably not `/etc/*`.  Installation of the completion functions may not be possible with the `builtin` build back-end.

As a last resort, you can always install it as part of the `modules` section, and let the user manually copy the files to the appropriate places, adding some instructions/examples in the README or gh-pages on how to do this. Something like:
```
$ sudo cp /usr/share/lua/5.1/busted/completions/bash/busted.bash /etc/bash_completion.d/`
$ sudo cp /usr/share/lua/5.1/busted/completions/zsh/_busted /usr/share/zsh/functions/`
```
Replacing `/usr/share/lua/5.1/` with whatever the `luarocks` lua install path is.

PR #317, opened 2015-01-02
Fix misleading error messages

o-lim:
This fixes the case where the file, line number, and error message reported are incorrect, showing the error message for a previous call to `error()` that was caught with a `pcall()`.

```lua
describe('Test assert after pcall', function()
  it('should report line number of assert', function()
    local ok, msg = pcall(function() error('error message') end)
    assert(false, 'should report this line number')
  end)
end)
```

In the above example, the error message reported would be from the `pcall()`, instead of the error message from the `assert()`, where the actual test failure occurred.

The solution is to not replace `error` in the test environment. This can lead to more inconsistent error messages that may be difficult to trace.  Go back to rewriting the error message to insert the file and line
information for complex error objects.


spec/cl_spec.lua
```diff
@@ -349,6 +349,24 @@ describe('Tests error messages through the command line', function()
     error_end()
   end)
 
+  it('when throwning an error table with __tostring', function()
```
DorianGray: typo!

spec/cl_spec.lua
```diff
@@ -349,6 +349,24 @@ describe('Tests error messages through the command line', function()
...
+    error_start()
+    local result = run('bin/busted --output=plainTerminal --tags=tostring --pattern=cl_error_messages.lua$')
+    local errmsg = result:match('\n(%./spec/.-)\n')
+    local expected = './spec/cl_error_messages.lua:17: {}'
+    assert.is_equal(expected, errmsg)
+    error_end()
+  end)
+
+  it('when throwning after a pcall', function()
```
DorianGray: typo!

DorianGray:
Typos aside, this looks really good. Will test and roll a new rc tonight.

DorianGray:
I especially like moving nil handling into rewrite message, a much more robust approach.

o-lim:
Just did a rebase to fix the typo.  Good catch.

DorianGray:
I was able to test this early. Good to go!

PR #321, opened 2015-01-06
Localize luassert to busted contexts

o-lim:
The `luassert` variables `assert`/`spy`/`mock`/`stub` should only be visible inside a busted context. Other modules outside of busted should be free to reuse/reassign these variable names without interfering with the test environment (and vice versa). If other modules need to use `luassert`, they should explicitly `require` it. Otherwise, `assert` should default to the builtin Lua `assert`.

This PR requires an update to `luassert` (Olivine-Labs/luassert#73) to fix a missing `require`, so the rockspec was also updated to specify luassert version 1.7.4 when it is released.


ajacksified:
Rebased - closed with #fa528e6101579705ba13ca20504046932dbc5cd2

Issue #320, opened 2015-01-05
access spec declaration methods from modules


leafo:
Sometimes I make helper modules with utility functions for declaring tests that are very similar:

```lua
-- test_helpers.lua
local function make_test(a, b)
  return it("should check " .. a .. " same as " .. b, function()
    assert.same(a, b)
  end)
end
return {
  make_test = make_test
}
```

And then use them like this: 
```lua
local make_test = require("test_helpers").make_test
return describe("things", function()
  make_test(1, 2)
  make_test(4, 4)
end)
```

In busted 2.0 this is no longer possible since the spec declaration methods like `it` and `describe` are only available within the `setfenv` functions.

One possible solution would be to make spec methods importable from the busted module: 

```lua
-- test_helpers.lua
local it = require("busted").it

local function()
  it("should do something", ...)
end
```

DorianGray:
This duplicates https://github.com/Olivine-Labs/busted/issues/319

DorianGray:
I made this the main issue since the explanation is better.

To solve this, we need to refactor init.lua to have one file per function, or set of functions. One would then require a specific function in to use it outside of files directly executed by the test runner.

IE::
local it = require 'busted.it'

Because we would have already required busted.it in busted init, the require inside of this external file would be fed the same instances of the functions that the init process is, therefore allowing them to operate on the active instance of busted.

o-lim:
I took a crack at this in PR #322.  Let me know what you think.

PR #322, opened 2015-01-07
Access spec declaration methods

o-lim:
This address issue #320, allowing helper modules to declare tests like:
```lua
-- test_helper.lua
local it = require 'busted'.it

local function make_test(a, b)
  it('should check ' .. a .. ' same as ' .. b, function()
    assert.same(a, b)
  end)
end

return {
  make_test = make_test
}
```
And then use the helper modules like:
```lua
-- test_spec.lua
local make_test = require 'test_helpers'.make_test
describe('things', function()
  make_test(1, 2)
  make_test(4, 4)
end)
```
This is done by exposing the busted executor functions via `require 'busted'` ~~or `require 'busted.executors'`~~ so they can be imported by helper modules.

### DIFF
--- a/busted/runner.lua
+++ b/busted/runner.lua
@@ -52,7 +52,7 @@ return function(options)
 
   local luacov = require 'busted.modules.luacov'()
 
-  require 'busted.init'(busted)
+  require 'busted'(busted)
 
   -- Default cli arg values
   local defaultOutput = path.is_windows and 'plainTerminal' or 'utfTerminal'
--- a/busted.lua
+++ b/busted.lua
@@ -1,0 +1,3 @@
+-- This is a dummy file so it can be used in busted's specs
+-- without adding ./?/init.lua to the lua path
+return require 'busted.init'
--- a/busted/init.lua
+++ b/busted/init.lua
@@ -13,15 +13,15 @@ local function shuffle(t, seed)
   return t
 end
 
-return function(busted)
-  local function remove(descriptors, element)
-    for _, descriptor in ipairs(descriptors) do
-      element.env[descriptor] = function(...)
-        error("'" .. descriptor .. "' not supported inside current context block", 2)
-      end
+local function remove(descriptors, element)
+  for _, descriptor in ipairs(descriptors) do
+    element.env[descriptor] = function(...)
+      error("'" .. descriptor .. "' not supported inside current context block", 2)
     end
   end
+end
 
+local function init(busted)
   local function exec(descriptor, element)
     if not element.env then element.env = {} end
 
@@ -201,3 +201,15 @@ return function(busted)
 
   return busted
 end
+
+return setmetatable({}, {
+  __call = function(self, busted)
+    init(busted)
+
+    return setmetatable(self, {
+      __index = function(self, descriptor)
+        return busted.executors[descriptor]
+      end
+    })
+  end
+})
--- a/busted/init.lua
+++ b/busted/init.lua
@@ -209,6 +209,10 @@ return setmetatable({}, {
     return setmetatable(self, {
       __index = function(self, descriptor)
         return busted.executors[descriptor]
+      end,
+
+      __newindex = function(self, key, value)
+        error('Attempt to modify busted')
       end
     })
   end
--- a/spec/executors_spec.lua
+++ b/spec/executors_spec.lua
@@ -37,4 +37,15 @@ describe('tests require "busted"', function()
     assert.is_equal(before_each, require 'busted'.before_each)
     assert.is_equal(after_each, require 'busted'.after_each)
   end)
+
+  it('functions cannot be overwritten', function()
+    local foo = function() assert(false) end
+    assert.has_error(function() require 'busted'.it = foo end)
+    assert.is_equal(it, require 'busted'.it)
+  end)
+
+  it('cannot add new fields', function()
+    local bar = function() assert(false) end
+    assert.has_error(function() require 'busted'.foo = bar end)
+  end)
 end)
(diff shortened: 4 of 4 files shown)


o-lim:
Couldn't really find an easy way to refactor `init.lua` to have one file per function.  Each function in `init.lua` needs the `busted` variable, so creating separate files per function would look something like:
```lua
-- init.lua
return function(busted)
  ...
  require 'busted.describe'(busted)
  require 'busted.it'(busted)
  ...
end
```
```lua
-- it.lua
return setmetatable({}, {
  __call = function(self, busted)
    local function it(element)
     ...
    end
    busted.register('it', it)
    setmetatable(self, {
      __call = function(self, ...) busted.executors.it(...) end
    })
  end
})
```
I tried a different approach to augment `init.lua` to return a table with the `busted.executors`
```lua
return setmetatable({}, {
  __call = function(self, busted)
    init(busted)

    return setmetatable(self, {
      __index = function(self, descriptor)
        return rawget(self, descriptor) or busted.executors[descriptor]
      end,
      __newindex = function(self, key, value)
        if busted.executors[key] then
          error("Cannot modify busted executor '" .. key .. "'")
        else
          rawset(self, key, value)
        end
    })
  end
})
```
So the busted runner would do `require 'busted.init'(busted)`, and helper modules could do `require 'busted'.it`.

I wasn't sure whether or not to expose the busted root context environment (via `busted.context.get('env')`) or just the executor functions.  Using the busted environment, you could also expose other functions/variables that are not registered via `busted.register`, and helper modules could potentially add functions/variables to the busted environment (that might be asking for trouble though).

DorianGray:
I've tested this briefly, looks good!

@leafo - what do you think?

iggyvolz:
You might want to squash your commits so they don't clog up the logs - but that is the developers' decisions (I know many large-scale projects such as homebrew or homebrew-cask will only accept squashed merge commits).  I can lead you through the process if you would like.

DorianGray:
I'd prefer not squashing. Lost history is worse than clogged logs.

I'd prefer if we were a little more strict about making sure every commit is a working commit, but honestly we don't have enough contributors for it to cause a problem. Regardless, please open a separate issue if you feel strongly about it.

leafo:
Appears to work.

This tripped me up though: 

```moon
print(require("busted").it) --> nil
print(require("busted.init").it) --> func
```

It's because in your rockspec you don't define a busted module, but only a `busted.init` module: 
```
    ['busted.init']                           = 'busted/init.lua',
```

I don't know if that's intentional but I would I typically do something like this: 
```
    ['busted']                           = 'busted/init.lua',
```

o-lim:
@leafo using `require ("busted").it` should still work because the lua path should have something like `/usr/share/lua/5.1/?/init.lua`.  If you are missing a `.../?/init.lua` in your lua path, then you will get that error.  You can also do `require ("busted.executors").it` as an alternative.  I had to do this in the spec files since `./?/init.lua` is not in the default lua path.

leafo:
@o-lim not quite true, luarocks changes how loading works. You'll actually get an empty module:

```
Lua 5.1.5  Copyright (C) 1994-2012 Lua.org, PUC-Rio
> for k,v in pairs(require("busted")) do print(k,v) end

(no output)
> print(package.path)
;./?.lua;/usr/share/lua/5.1/?.lua;/usr/share/lua/5.1/?/init.lua;/usr/lib/lua/5.1/?.lua;/usr/lib/lua/5.1/?/init.lua;/home/leafo/.luarocks/share/lua/5.1/?.lua;/home/leafo/.luarocks/share/lua/5.1/?/init.lua
```

```bash
$ file /home/leafo/.luarocks/share/lua/5.1/busted/init.lua
/home/leafo/.luarocks/share/lua/5.1/busted/init.lua: ASCII text
```

leafo:
(I wrote something else but I'm rivising it :))

I don't get an error when I `require "busted"` I get a version of the module that doesn't seem to work correctly. Only briefly looked at it, not sure what the cause is.

Regardless, I think the preferred way is to not reference init in your module.

o-lim:
I think I understand the problem you are running into.  Since the strings `"busted"` and `"busted.init"` are different, Lua thinks they are two separate modules, and so loads both of them independently of each other.  The two statements `require "busted"` and `require "busted.init"` really refer to the same file, so I just assumed that Lua would also treat them as aliases for one another.

I just rebased and pushed a fix so that `require 'busted'` won't reload the file.

busted/runner.lua
```diff
@@ -52,7 +52,7 @@ return function(options)
 
   local luacov = require 'busted.modules.luacov'()
 
-  require 'busted.init'(busted)
+  package.loaded['busted'] = require 'busted.init'(busted)
```
leafo: I don't think I agree with this solution. Why would you do it in `runner.lua`? Doesn't it make sense to do it in the module that it affects, `init.lua`. Personally I think what I suggested in the issue is a better approach. It should not be possible to load a copy of the same module.

o-lim:
I pushed another update that I think should address your concerns.

o-lim:
I made another update.  Simplified the implementation by making the `busted` table strictly read-only.

leafo:
This happens to me with your latest changes:

```
Lua 5.1.5  Copyright (C) 1994-2012 Lua.org, PUC-Rio
> print(require("busted").it)
stdin:1: attempt to index a function value
stack traceback:
	stdin:1: in main chunk
	[C]: ?
```

leafo:
Nevermind post I just deleted, everything looks good here! (Mostly, I notice that it's about 60% slower running this suite than the 1.11 I was running before, but unrelated to this issue)

PR #328, opened 2015-02-06
Support loading output handler via Lua search path

o-lim:
This allows an output handler to be loaded from a module found in the Lua search path (or path specified by the --lpath option). The output handler can still be loaded from a lua or moonscript file as well. If the output handler specified on the command line cannot be found in `busted/outputHandlers/`, then attempt to load
the output handler as if it were in the Lua module search path.

### DIFF
--- a/.travis_setup.sh
+++ b/.travis_setup.sh
@@ -24,8 +24,8 @@ else
 fi
 
 cd ..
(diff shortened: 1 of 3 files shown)


o-lim:
Updated PR to rebase on top of PR #331 so the travis build would work.

Issue #327, opened 2015-02-03
busted --lpath isn't functioning


TheCycoONE:
As can be seen in our pull request https://github.com/CorsixTH/CorsixTH/pull/661, busted --lpath isn't functioning installed in a clean (TravisCI) environment with luarocks.

A workaround is to prepend the path to LUA_PATH manually.

working:
luafilesystem 1.6.2-2
busted 2.0.rc3-0

non-working:
luafilesystem 1.6.3-1
busted 2.0.rc5-0

o-lim:
Fixed in PR #329 .  However, it was technically not working in busted 2.0.rc3-0 either.  I looked more closely at your repo, and found that you have a `.busted` file with a typo (backslash instead of forward slash).
```lua
return {
  default = {
    lpath = "../Lua\?.lua",
  },
}
```
when it should be
```lua
return {
  default = {
    lpath = "../Lua/?.lua",
  },
}
```
The error was due to the way the `.busted` file was being loaded.  The `--lpath` command-line argument was being overridden with the value in the `.busted` file, when it should have been the other way around.  The command-line arguments should override the `.busted` file.

If you fix your `.busted` file, you shouldn't need to use `--lpath` in your rockspec.

TheCycoONE:
Thanks for being so thorough @o-lim!

PR #329, opened 2015-02-06
Fix issue #327 busted --lpath isn't functioning 

o-lim:
This fixes issue #327
- The default `.busted` task was overriding the command-line argument for `--lpath`


o-lim:
Updated PR to rebase on top of PR #331 so the travis build would work.

PR #335, opened 2015-02-16
Access for busted root context environment variables

o-lim:
This allows helper modules to use variables/functions declared in the busted environment, such as `assert`, `spy`, `mock`, `stub`, etc. This complements the spec declarations methods (`describe`, `it`, etc.), which can already be used by helper modules.


ajacksified:
Nice!

PR #336, opened 2015-02-16
Default output handler depends on stdout is a tty

o-lim:
Set default output handler to `utfTerminal` if `stdout` is a tty, otherwise set the default output handler to `plainTerminal`.

This allows file redirection to automatically default to `plainTerminal`.  For example
```
$ busted > log.txt
```
will use `plainTerminal`, while
```
$ busted
```
will use `utfTerminal`.

In addition, there is no need to have any special logic for Windows since the `ansicolors` module already takes care of this by removing any ansi color codes when running on Windows, only enabling ansi color codes when ansicon is present.


ajacksified:
:fireworks:

PR #337, opened 2015-02-16
Fix tests to work on Windows

o-lim:
This fixes `cl_spec.lua` so that the tests will run and pass on Windows.

Test on Windows 7 with the following shells:
* MinGW with bash
* cmd.exe


ajacksified:
I don't have a Windows machine handy, but I'll take your word for it.

PR #338, opened 2015-02-16
Add --no-recurse flag

o-lim:
This allows busted to run tests in a directory without recursing into any sub-directories.

This is on top of PR #337 so that tests will still pass when run on Windows.


ajacksified:
Good idea. Thanks! I'll add it to the docs once it's released.

PR #340, opened 2015-02-17
Close Lua state on exit

o-lim:
Calls `os.exit` with `close=true` to allow Lua to close the Lua state before exiting.


ajacksified:
:balloon:

PR #344, opened 2015-02-18
Fix #333: junit dies when trace is nil

### LINKED ISSUE #333: junit dies when trace is nil
I'm getting this during a build:

```text
00:42:42,594 WARN  - /usr/home/quickbuild/buildagent/workspace/root/neovim/integrate/.deps/usr/bin/luajit: .../.deps/usr/share/lua/5.1/busted/outputHandlers/junit.lua:59: attempt to index global 'trace' (a nil value)
00:42:42,594 WARN  - stack traceback:
00:42:42,594 WARN  -     .../.deps/usr/share/lua/5.1/busted/outputHandlers/junit.lua:59: in function 'fn'
00:42:42,594 WARN  -     ...ot/neovim/integrate/.deps/usr/share/lua/5.1/mediator.lua:98: in function 'publish'
00:42:42,594 WARN  -     ...neovim/integrate/.deps/usr/share/lua/5.1/busted/core.lua:143: in function 'safe'
00:42:42,594 WARN  -     ...neovim/integrate/.deps/usr/share/lua/5.1/busted/core.lua:199: in function 'execute'
00:42:42,594 WARN  -     ./.deps/usr/lib/luarocks/rocks/busted/2.0.rc4-0/bin/busted:220: in main chunk
00:42:42,594 WARN  -     [C]: at 0x00404180
```

It comes from this function in junit.lua:

```lua
  handler.errorFile = function(element, parent, message, trace)
    xml_doc.attr.errors = xml_doc.attr.errors + 1
    xml_doc:addtag('error')
    xml_doc:text(trace.traceback)
    xml_doc:up()

    return nil, true
  end
```

It appears that trace can be nil sometimes, and we're dereferencing it here.  So it appears this would be the proper thing to do:

```lua
  handler.errorFile = function(element, parent, message, trace)
    xml_doc.attr.errors = xml_doc.attr.errors + 1
    xml_doc:addtag('error')
    if trace then
      xml_doc:text(trace.traceback)
    end
    xml_doc:up()

    return nil, true
  end
```

That would avoid this issue, but it would leave the user wondering what went wrong.  Should the `message` parameter be used somewhere in here?  Would you take a patch with the above fix?

jszakmeister:
This also adds message to the output, since it may be the only
information we have if trace is nil.

### DIFF
--- a/busted/outputHandlers/junit.lua
+++ b/busted/outputHandlers/junit.lua
@@ -54,7 +54,10 @@ return function(options, busted)
   handler.errorFile = function(element, parent, message, trace)
     xml_doc.attr.errors = xml_doc.attr.errors + 1
     xml_doc:addtag('error')
-    xml_doc:text(trace.traceback)
+    xml_doc:text(message)
+    if trace then
+      xml_doc:text(trace.traceback)
+    end
     xml_doc:up()
 
     return nil, true


ajacksified:
Can you verify that latest master, with https://github.com/Olivine-Labs/busted/pull/341, fixes this for you as well?

jszakmeister:
@ajacksified I tried out the new version on master, and it does fix the issue.  And thank you for adding the extra information into the failure messages!  That will be most helpful.

I assume you don't need this anymore, so I'll close this PR.

PR #347, opened 2015-02-19
Cleanup output handler loader and file loaders

o-lim:
* Simplify moonscript and terra file loaders
* Cleanup error handling in output handler loader
* Add support for loading moonscript output handlers


ajacksified:
:fireworks:

PR #350, opened 2015-02-20
Add a timestamp to the testsuite tag in the jUnit test report.

jszakmeister:


jszakmeister:
The time here is reported in UTC.  Do you think it better to use the local time?

o-lim:
Maybe local time with a UTC offset would be better?

o-lim:
What about something like
```lua
os.date('%Y-%m-%d %T %Z %z')
```
which would look like
```
2015-02-20 13:10:35 PST -0800
```

jszakmeister:
The problem is that there's an underlying scheme that we should adhere to, and it's not clear that ISO8601 with timezone is supported.  Looking at [this](http://help.catchsoftware.com/display/ET/JUnit+Format) suggests no, but looking at the actual report generator may say otherwise.  The hope is that tools like Jenkins and QuickBuild can parse the format and show it as part of the report.

o-lim:
In that case, it's probably better to leave it as UTC.

DorianGray:
Is this good to merge? Or are more changes necessary?

o-lim:
I think it is good to merge.  What do you think @jszakmeister?

jszakmeister:
Yes.  Ant produces junit files with the timestamp set to UTC, so I think that's what we should use here.

PR #353, opened 2015-02-23
Fix output handlers

o-lim:
Updated output handlers
* Make sure base output handler is run before derived handlers
* Fix output for multiple runs
* Fix junit output to better match junit xml format
 * Mark pending tests as `<skipped/>`
 * Nest  errors that occur in a test inside `<testcase>`
 * Nest test suite runs inside `<testsuites>`


DorianGray:
Mind rebasing?

PR #355, opened 2015-02-23
Helper script loader

o-lim:
Update helper script loading
* Support for moonscript helpers
* Support loading helper from a library module in `package.path`
* Output error message when helper fails to load
* Add command-line option to forward options to helper script


DorianGray:
Mind rebasing?

PR #358, opened 2015-02-23
Fix duplicate parent variable

o-lim:


DorianGray:
Funny, I even noticed this and then forgot and clicked merge anyways =(

Issue #332, opened 2015-02-09
call os.exit with close=true


starius:
I run busted tests of my C/C++ module under valgrind. Because busted calls os.exit(code), Lua state is not closed, destructors of some objects are not called, which looks like memory leak for valgrind. I have an idea how to fix this. Lua >= 5.2 has optional argument "close" of function os.exit, which tells Lua to close Lua state before exiting. On Lua 5.1, workaround is calling collectgarbage() three times. before os.exit().

*UPD*. Another option for Lua 5.1 is not calling os.exit() at all, if exit code is success.

PR #359, opened 2015-02-23
fix memoryleak in userdata on os.exit in Lua 5.1

starius:
see #332

### DIFF
--- a/busted/compatibility.lua
+++ b/busted/compatibility.lua
@@ -30,5 +30,31 @@ return {
     if f ~= 0 then return f end
   end,
 
-  unpack = table.unpack or unpack
+  unpack = table.unpack or unpack,
+
+  osexit = function(code, close)
+    if close and _VERSION == 'Lua 5.1' then
+      -- From Lua 5.1 manual:
+      -- > The userdata itself is freed only in the next
+      -- > garbage-collection cycle.
+      -- Call collectgarbage() while collectgarbage('count')
+      -- changes + 3 times, at least 3 times,
+      -- at max 100 times (to prevent infinite loop).
+      local times_const = 0
+      for i = 1, 100 do
+        local count_before = collectgarbage("count")
+        collectgarbage()
+        local count_after = collectgarbage("count")
+        if count_after == count_before then
+          times_const = times_const + 1
+          if times_const > 3 then
+            break
+          end
+        else
+          times_const = 0
+        end
+      end
+    end
+    os.exit(code, close)
+  end,
 }
--- a/busted/runner.lua
+++ b/busted/runner.lua
@@ -19,6 +19,8 @@ return function(options)
 
   local luacov = require 'busted.modules.luacov'()
 
+  local osexit = require 'busted.compatibility'.osexit
+
   require 'busted'(busted)
 
   -- Default cli arg values
@@ -80,13 +82,13 @@ return function(options)
   -- Parse the cli arguments
   local cliArgs = cli:parse(arg)
   if not cliArgs then
-    os.exit(1, true)
+    osexit(1, true)
   end
 
   -- Return early if only asked for the version
   if cliArgs.version then
     print(busted.version)
-    os.exit(0, true)
+    osexit(0, true)
   end
 
   -- Load current working directory
@@ -152,7 +154,7 @@ return function(options)
     for _, included in pairs(tags) do
       if excluded == included then
         print('Cannot use --tags and --exclude-tags for the same tags')
-        os.exit(1, true)
+        osexit(1, true)
       end
     end
   end
@@ -381,5 +383,5 @@ return function(options)
       exit = 255
     end
   end
-  os.exit(exit, true)
+  osexit(exit, true)
 end


o-lim:
The problem with not calling `os.exit` when the exit code is zero, is when running in standalone mode the test file will be executed twice.

Running standalone, you can do
```
$ lua myspec.lua
```
instead of 
```
$ busted myspec.lua
```
If I had a file mypsec.lua
```lua
require 'busted.runner'()

print('Running test . . .')
describe('Test standalone', function()
  it('test always passes', function()
  end)
end)
```
Running standalone with this PR
```
$ lua myspec.lua
Running tests . . .
●
1 success / 0 failures / 0 errors / 0 pending : 0.0 seconds
Running tests . . .
```
instead of
```
$ lua myspec.lua
Running tests . . .
●
1 success / 0 failures / 0 errors / 0 pending : 0.0 seconds
```
and running with busted explicitly runs as expected
```
$ busted myspec.lua
Running tests . . .
●
1 success / 0 failures / 0 errors / 0 pending : 0.0 seconds
```

starius:
Thanks!

There are two options I want to discuss.

1) Call `os.exit` even if exit code is zero. Calling `collectgarbage` before `os.exit` prevents memory leak if all userdata are collected. This condition may be unsatisfied in standalone mode (or may not?).

2) Change other code in busted given the fact `os.exit` is not called if no error occurs. Calling `os.exit` has to be isolated from `runner` in this case. This is larger change, but it makes busted more clean.

o-lim:
I believe `os.exit` is only called from `runner`.

Would it be sufficient to do some like:
```lua
local function exit(code, close)
  if  close and _VERSION == 'Lua 5.1' then
    collectgarbage()
    collectgarbage()
    collectgarbage()
  end
  os.exit(code, close)
end
```
and modify `runner` to call `exit` instead of `os.exit`?  Can you verify whether or not this works for you if your run in standalone mode?

starius:
This approach works well.
* [x] No memory leaks are reported by valgrind (Lua 5.1).
* [x] Code is executed once in standalone mode (tested on file `myspec.lua`)

The patch proposed:
```diff
From aa3ee0bf606c9a46465e959f52ef53583cb3a5ea Mon Sep 17 00:00:00 2001
From: Boris Nagaev <bnagaev@gmail.com>
Date: Tue, 24 Feb 2015 00:23:38 +0300
Subject: [PATCH] call os.exit even if exit code is zero

Not calling os.exit results in running a code
twice in standalone mode.

If Lua version is 5.1 and close=true,
collectgarbage() is called 3 times before os.exit.
---
 busted/runner.lua |   26 +++++++++++++-------------
 1 file changed, 13 insertions(+), 13 deletions(-)

diff --git a/busted/runner.lua b/busted/runner.lua
index 60c73aa..6b573f1 100644
--- a/busted/runner.lua
+++ b/busted/runner.lua
@@ -5,6 +5,15 @@ local term = require 'term'
 local utils = require 'busted.utils'
 local loaded = false
 
+local function osexit(code, close)
+  if close and _VERSION == 'Lua 5.1' then
+    collectgarbage()
+    collectgarbage()
+    collectgarbage()
+  end
+  os.exit(code, close)
+end
+
 return function(options)
   if loaded then return else loaded = true end
 
@@ -80,13 +89,13 @@ return function(options)
   -- Parse the cli arguments
   local cliArgs = cli:parse(arg)
   if not cliArgs then
-    os.exit(1, true)
+    osexit(1, true)
   end
 
   -- Return early if only asked for the version
   if cliArgs.version then
     print(busted.version)
-    os.exit(0, true)
+    osexit(0, true)
   end
 
   -- Load current working directory
@@ -152,7 +161,7 @@ return function(options)
     for _, included in pairs(tags) do
       if excluded == included then
         print('Cannot use --tags and --exclude-tags for the same tags')
-        os.exit(1, true)
+        osexit(1, true)
       end
     end
   end
@@ -381,14 +390,5 @@ return function(options)
       exit = 255
     end
   end
-  if exit ~= 0 then
-    -- if exit == 0, calling os.exit is not needed.
-    -- os.exit can cause memoryleak in userdata.
-    -- collectgarbage called before os.exit can prevent
-    -- memoryleak in userdata.
-    collectgarbage()
-    collectgarbage()
-    collectgarbage()
-    os.exit(exit, true)
-  end
+  osexit(exit, true)
 end
-- 
1.7.10.4
```

DorianGray:
I'd prefer if this was moved into the compatibility module. It would be nice if all os/vm specific code resided there.

starius:
Great idea!

However I still need an option not to call `os.exit` on success, because valgrind marks all Lua 5.1 memory blocks as "still reachable":
```
still reachable: 2,660,638 bytes in 28,231 blocks
```
Even if memory leak occurs, it is counted as "still reachable", which is not an error. This is caused by os.exit (Lua 5.1).

I think, this problem can be solved with command line option preventing `os.exit` on success.

o-lim:
@starius: Just as a general question, why is it necessary to call `collectgarbage()` 3 times?  Why isn't it sufficient to call it just once?

starius:
@o-lim this is part of Lua 5.1 manual:
> At the end of each garbage-collection cycle, the finalizers for userdata are called in reverse order of their creation, among those collected in that cycle. That is, the first finalizer to be called is the one associated with the userdata created last in the program. The userdata itself is freed only in the next garbage-collection cycle.

If collecting userdata A causes collecting another userdata B, then userdata B is freed only in the next to next garbage-collection cycle (i.e., third). By the way, I have no idea, what happens, if collecting userdata B causes collecting userdata C. Probably, a memory leak occurs. In this case garbage collecting has to be run more than 3 times. The following (not tested) workaround can be applied. Run `collectgarbage` at least 3 times while memory in use changes + 3 times.
```lua
local times_const = 0
while true do
  local count_before = collectgarbage("count")
  collectgarbage()
  local count_after = collectgarbage("count")
  if count_after == count_before then
    times_const = times_const + 1
    if times_const > 3 then
      break
    end
  else
    times_const = 0
  end
end
```
Please correct me if I am in error about this garbage collecting issues.

o-lim:
Thanks for the explanation.

DorianGray:
Your snippet looks good. If you roll all of these things together and rebase, I'd be happy to merge!

starius:
I think, `while true` should be replaced with `for i = 1, N` with large enough N to prevent infinite loop here.

DorianGray:
As long as the lua garbage collection internals are working properly I don't believe that will ever happen but I agree that having an extra layer of protection is useful.

DorianGray:
:+1:

PR #362, opened 2015-02-24
Insert junit timestamp in each test suite

o-lim:
Insert timestamp in each test suite header instead of the overall header for all test suites.


mpeterv:
It looks like local declaration of testStartTime was lost at some point..?

o-lim:
You're right, not sure how that happened. I just put it back in.  Good catch.

Issue #356, opened 2015-02-23
Merge argument order looks wrong in configuration_loader.lua


geoffleyland:
tablex.merge preserves values from the table on the right when there's a conflict:

    $ lua
    > t = require"pl.tablex"
    > a=t.merge({a=1},{a=3},true)
    > =a.a
    3

But lines 17 and 25 of configuration_loader.lua have the default configuration on the right, and the config file on the left, so I suspect the config file can't overwrite anything.  Swapping them around fixed my problem anyway.

Looking further, it would appear that fixing #327 caused this.  As far as I can tell, at this point, config contains both default values and command-line values.  So it might be more a question of the order in which default, .busted and command-line values are merged, rather that swapping the order here.

Unfortunately I'm not sure you can tell the difference between what cliargs really read on the command line, and what it added for you as a default.

o-lim:
I don't think the configuration file is supposed to overwrite anything.  The way I understand it is that, the command line is supposed to overwrite what is in the configuration file.

Quoting from http://olivinelabs.com/busted:
> Busted 1.6 added a concept of "tasks", or predefined busted configuration options. You can create a .busted file in the root, which is automatically loaded if it exists. Default options are run if no task is specified. You can add any argument available in the CLI (listed above), using the long name. (Use verbose = true, not v = true.) Any arguments you specify will override those set in the task.

geoffleyland:
Yes, but the default arguments come through at the same time as the commands that were actually put on the command line.

For example, runner.lua has

    local lpathprefix = './src/?.lua;./src/?/?.lua;./src/?/init.lua'

Then

    cli:add_option('-m, --lpath=PATH', 'optional path to be prefixed to the Lua module search path', lpathprefix)

So by the time you get to

    local config, err = configLoader(configFile, cliArgs)

even if you only type `busted` on the command line, `cliArgs` contains a value for `lpath`.

So if you have a `.busted` with

    return { default = { lpath = "./my_path/?/?.lua" } }

It'll be ignored. (Try it!)

The problem isn't that things that are actually on the command-line are overridden, it's that the default values look like they came from the command-line.

DorianGray:
tl;dr; If there is a .busted file, we need to -not- default cliargs.

geoffleyland:
Is it as simple as that?  You might want some the other default args that are neither on the cli nor in .busted, and also, .busted is searched for in --cwd.
[...]

PR #363, opened 2015-02-24
Fix configuration loading

o-lim:
This updates the configuration loader to merge command-line options with a specified configuration task, taking in order of precedence:
1. Options specified on the command-line
2. Options specified in the configuration file
3. Built-in command-line option defaults

Also, exit with error when there is an error loading the configuration.

Fix issue #356


DorianGray:
This looks great!

PR #365, opened 2015-02-26
Update command-line options

o-lim:
* Add `-l` options as an alias for `--list`
* Rename `--no-recurse` to `--no-recursive`
 * This spelling matches other cli's which have a `--recursive` option (i.e. grep, cp, chmod, etc)


ajacksified:
:+1:

PR #371, opened 2015-03-05
Refactors and improvements

o-lim:
Made the following changes:
* Refactor CLI setup, validation, parsing into a separate file
* Refactor test filtering into a separate file
* Refactor file/describe block execution
* Refactor error/failure output
* Only publish test start/end if `before_each` passes
* Do not execute `finally` block if test is not executed
* Allow list type options to be specified multiple times
* Support multiple root file paths to be specified on the command-line
* Support multiple `--cwd` options, interpreting each one relative to the previous one
* Support multiple `--filter` and `--filter-out` options
* Export `publish`/`subscribe` to external modules
* Add `unsubscribe` function
* Execute `setup`/`teardown` for test suite


busted/modules/filter_loader.lua
```diff
@@ -0,0 +1,95 @@
+
```
ajacksified: nit: newline

o-lim:
@mpeterv: Thanks catching that.  I fixed it and added some tests for the configuration loading.

DorianGray:
:100: This is fantastic. I love the cleanup and the new tests! Especially the init and cli breakouts. I'll do some deeper testing tonight, but the code looks pristine.

DorianGray:
:+1:

PR #373, opened 2015-03-06
Replace ansicolors with lua-term

mpeterv:
Since lua-term is already used and has a submodule equivalent to ansicolors,
use lua-term for color handling. Removed ansicolors dependency from rockspec.


ajacksified:
Less dependencies \o/

:+1:

mpeterv:
I'm sorry, apparently lua-term doesn't disable coloring on windows like ansicolors. I'll send a PR with a fix.

PR #374, opened 2015-03-06
Fix passing arguments to helpers

mpeterv:
Runner currently passes array with an empty string to helpers if no arguments are provided using -X\<helper\> options. This causes error when helper calls cli_args:parse on args. The issue is that when `cliArgs.Xoutput` is nil, `utils.split(cliArgs.Xoutput or '', ',') or {}` == `utils.split('', ',') or {}` == `{''} or {}` == `{''}`.


ajacksified:
:+1:

PR #376, opened 2015-03-10
Do not use colors on Windows if ANSICON is not defined

mpeterv:
This fixes an issue with #373: unlike ansicolors, lua-term does not disable colors on Windows where it is typically not supported.


DorianGray:
:+1:

PR #378, opened 2015-03-13
Don't execute setup/teardown on skipped suites

### LINKED ISSUE #377: setup/teardown hooks are always executed even if no tests are going to be executed for a suite
I frequently use the --tags option to focus on a single test while developing/debugging, but the setup hooks are always executed for all "describes", even when its children tests have been skipped by the tag filter.

Normally that wouldn't be an issue, but Neovim setup hooks are quite expensive, especially when running the tests with valgrind. Would it be possible to defer the setup/teardown hooks in such a way that they are only executed if the suite has at least one test that will be executed?

tarruda:
If a describe block contains no tests or only tests that have been skipped(due to --tags or --exclude-tags), don't execute its setup/teardown hooks. The goal is to fix #377 

I'm very unfamiliar with the code and this change was based on a quick skim so it may contain stupid mistakes. This was adapted from 2.0rc7(which is the version I have installed and was working for me)


tarruda:
`Tests helper script can add {setup,teardown} to test suite` failed. It seems hooks in helper scripts cant rely on recursively counting child `it` blocks. Any suggestions?

o-lim:
Looks like you are modifying the wrong file.  Furthermore, this approach will not work since you can't count `it` blocks in nested blocks.  See comment in #377.

tarruda:
@o-lim can you give it another review? I tried again with a different approach.

o-lim:
Like I said before, this approach does not work.  You cannot count the number of `it` blocks in any nested contexts.  This is because they are not added or discovered until the block is executed.  The only time you would have an accurate count of `it` blocks would be after `execute`, basically when you run `teardown`.  The information you want simply is not available at the time of `setup`, you only have partial information.  You can get the number of `it` blocks in the current context, but not any any of the nested context blocks.

As I mentioned in #377, the following example will not run `setup`.  The 2 `it` blocks `inner block1` and `inner block2` will not be added to the context until `inner block1` and `inner block2` are executed, respectively.
```lua
describe('outer block', function()
  setup(function()
  ...
  end)

  describe('inner block1', function()
    it('inner test1', function()
    ...
    end)
  end)

  describe('inner block2', function()
    it('inner test2', function()
    ...
    end)
  end)
end)
```

tarruda:
Ok I understand the problem now. I'm gonna give it another shot using the "once" approach

tarruda:
@o-lim I refactored so the setup hook uses the "once pattern". Clearly this doesn't work for teardown, so whenever a test is executed the test count for the current context and all parent contexts is increased. When the context is finished, the teardown hook will only be ran if at least one test was executed.

Also added a couple more tests to verify the behavior in nested suites.

tarruda:
This last commit is a bit more elegant since it doesn't rely on the "once" function hack. Instead it implements the "execAllOnce" method which achieves a similar effect. It also implements a generic `set_parents` method in the context to set a property to the current context and all parents, this is used to communicate callers up in the stack that the teardown hook should be executed.

I you prefer the new variant I will squash into the first.

tarruda:
Note sure why tests are failing on travis. In my local environment every test is passing

tarruda:
oh, I need to run `bin/busted` and not the installed version of busted :)

tarruda:
Ok, its definitely much simpler to use the "once" function wrapping trick, I reverted the `execAllOnce` commit. There's only one failing test now:

```
Error → ./spec/core_spec.lua @ 381
tests unsupported functions in setup/before_each/after_each/teardown setup
./spec/core_spec.lua:364: Expected a different error.
Caught:
(no error)
Expected:
(string) ''randomize' not supported inside current context block'
```

Not what should be done about this one since I dont really understand whats going on. Any ideas?

o-lim:
This looks more complicated than it needs to be.
- You don't need to count any tests
 - `teardown` should be run if and only if `setup` was run
 - Changes to `busted/context.lua` are not necessary
- I don't think `busted.register` should have any special code for `setup`/`teardown`
- I don't think you need to add any tests to `cl_spec.lua`, unless you need to test with specific command-line options
 - You are changing the core functionality of busted so you would need new tests in `core_spec.lua`
 - `cl_spec.lua`, I believe is used for testing busted execution with command-line options
- This won't work if you have `setup`/`teardown` in a helper script and `--repeat=2`
 - The second execution of the test suite will probably skip the `setup`/`teardown` in a helper script

tarruda:
> This looks more complicated than it needs to be.

I think that too, I will just refactor my tests to use `before_each(once(cb))` instead of `setup`. Sorry for the noise

o-lim:
I have some ideas about how to do this.  Let me see if I can find some time to try it out.

tarruda:
> I have some ideas about how to do this. Let me see if I can find some time to try it out.

Thanks. If you do, I will adapt my code

tarruda:
I have opened neovim/neovim#2146 to replace `setup` calls by `before_first` which uses the `once` hack mentioned in #377, but I wont merge it for the next few days so it would be great if you find some time to implement this before.

If this proves to be too hard or challenging(to get the tests passing for example), could you add a way to override the native `setup` or at least inject globals that make use of `before_each`? The only I have found to do it was by injecting this code at the top of each file:

```lua
before_first = function(cb) before_each(once(cb)) end
```

The reason is because I couldn't access the `before_each` function from the test helper.

PR #380, opened 2015-03-14
Restore arg after loading helper script

### LINKED ISSUE #379: Allow arg to be seen by unittests...
With all the clearing going on, I believe `arg` is being cleared too.  We currently make use of it in some of Neovim's testing and would like to continue to do so.

o-lim:
This should fix #379 when a helper script is used.


jszakmeister:
Fantastic!  Any chance of a new -rc release?

PR #382, opened 2015-03-16
Arrow output character is not ascii

nehz:
In windows 7 (not utf8 console), the arrow character appears as random letters.
Changed arrow character to ascii characters.


DorianGray:
Not sure why this has failing tests...

o-lim:
Some of the tests check the output strings.  If `→` changed to `=>`, then it would also have to be updated in the tests that are looking for `Error →` or `Pending →`.

nehz:
Yeah I can see for example in:
https://github.com/Olivine-Labs/busted/blob/master/spec/cl_spec.lua
`local _, numErrors = result:gsub('Error → .-\n','')`

Is it ok to go and change these test cases?

o-lim:
Yes, you should change the test to match so that they pass.

What do you think of using `->` for the arrow, to more closely match the single arrow `→` (versus `=>` matching `⇒`)?

nehz:
Changed to `->`. Updated tests.

PR #385, opened 2015-03-16
chdir to cwd

o-lim:
Rename refactor -d, --cwd options to -C, --directory, and update option to change to the specified directory before running any tests or doing anything else. This is consistent with other commands like make and tar.


DorianGray:
Did you mean to remove the lpath and cpath code? I don't see how it works without it.

o-lim:
Yes, I did mean to remove the lpath and cpath code.  They are no longer necessary because of the `chdir`.  After you `chdir(cwd)`, the `./` path is now your `cwd` path, so changing the lpath and cpath can be removed.

o-lim:
Also, the following is still in `runner.lua`
```lua
 -- Add additional package paths based on lpath and cpath cliArgs
if #cliArgs.lpath > 0 then
  package.path = (cliArgs.lpath .. ';' .. package.path):gsub(';;',';')
end
if #cliArgs.cpath > 0 then
  package.cpath = (cliArgs.cpath .. ';' .. package.cpath):gsub(';;',';')
end
```

DorianGray:
needs rebase =(

o-lim:
Just rebased.

PR #388, opened 2015-03-18
Add more tests for lazy/strict setup/teardown

o-lim:


tarruda:
Just curious: Is there a reason for not making lazy mode the default(or even the only) behavior? I imagine that programmers use `setup`/`teardown` hooks to perform expensive setup for a group of tests, and I don't see why anyone would want to have these hooks called if the tests are not executed.

o-lim:
It's possible that some users could be relying on the strict behavior of `setup`/`teardown`.  So making it default to strict maintains backwards compatibility.  Also, there may be cases where you would want to control the behavior of specific blocks to be strict or lazy.  For example, there is nothing stopping you from using an `assert` in `setup` or `teardown`, so effectively the `setup`/`teardown` becomes part of your test, and if you have no `it` blocks then `setup`/`teardown` is your test.

DorianGray:
We haven't released 2.0 yet! We can break compat for a better interface!

Issue #372, opened 2015-03-06
Make module reloading opt-in rather than opt-out...


jszakmeister:
I'm a bit worried about user experience here.  Mike Pall said that the [ffi module cannot be reload](https://www.freelists.org/post/luajit/Assertion-error,5), and I think he's right in that C-based modules may have a number of issues on this front too.

So rather than leave folks to tracking down obscure issues, perhaps it would be better to opt-in for the test isolation, especially since it's clear it doesn't work everywhere?

ajacksified:
Hmm, this is a tough one. I would think that more people will have issues with modules *not* reloading than will have issues *with* reloading, but on the other hand, when you do have issues, they're hard to track down. Thoughts, anyone else?

jszakmeister:
@ajacksified I see it the other way around.  Your modules don't normally reload in a Lua app, why would you expect them to in a Lua unit tester?  I believe in regular Lua would have issues with some of the C modules being reloaded.  It seems more like a hack than a supported feature of the language.

Separate processes would probably be the best way to get isolation, but Windows is painful on that front.  Not only do you not have fork(), the process startup time is 100x to 1000x times slower than Linux. :-(

mpeterv:
I'm not sure how sandboxing is done currently, I assume it is implemented in [busted/context.lua](https://github.com/Olivine-Labs/busted/blob/f186d640f05d8d84de576225bac1c7ea4b37350d/busted/context.lua) with `save` function called before loading a file and `restore` after? Is there any environment assigned to main chunks of files, or does it break if the file sets globals which are used in `restore` function (e.g. `_G` = nil)? Does removing lines 20-22 fix the issue?

mpeterv:
Also, I think I agree with @jszakmeister - if you want to test if _your_ module breaks because of reloading, that's easy to do explicitly; if a module you use breaks on reload (like `ffi`), it shouldn't be your problem; and if a module you use breaks because your module messes up with it, reloading it only hides the issue.

o-lim:
From the perspective of using a module in a Lua app, you would not expect any module to be reloaded.  However, when unit testing you want to be able start each test "clean" and not impacted by the success or failure of a previous test.  Since many Lua modules will modify their internal state, the only way to achieve test isolation is to reload the module being tested or run each test in a separate process.  Running each test in either scenario comes with it's own set of hurdles to overcome.  

[...]

PR #389, opened 2015-03-19
Add support for insulate/expose

o-lim:
Addresses issue #372

This adds `describe` aliases that allow one to `insulate` the test environment or `expose` the test environment to outer context blocks.

Test environment insulation saves the global table `_G` and the currently loaded packages `package.loaded`, restoring them to their original state at the completion of an `insulate` block.

Exposing a test environment exports any changes made to `_G` and `package.loaded` to subsequent context blocks. In addition, any global variables created inside an `expose` block are created in the
environment of the context block 2 levels out. Using `expose` at the root of a file will promote any requires and globals to the root environment, which will then spillover into subsequent tests.

The `insulate` and `expose` blocks can be nested inside other blocks, as well as nested inside of each other. This provides full control of the level of isolation that may be desired.

By default all file contexts are insulated.  The command-line flag `--no-auto-insulate` can be used to disable file context insulation.  In this case, all test insulation would be explicit.


DorianGray:
:100:

tarruda:
:+1: 

hoping for a new release soon :smile:

o-lim:
Don't forget to update the version string.

Issue #375, opened 2015-03-09
Create the output handler that prints at least test name immediately test is run


ZyX-I:
All currently used output handlers assume that busted has a chance to shut down properly. This is not the case if the test is triggering lua(jit) bug, uses buggy C modules or has bug in code that uses luajit cffi. The only output handlers that give any information in this case are utfTerminal and plainTerminal, but counting `+` signs while you don’t know the order of execution of tests is not fun.

I suggest that there must be an output handler that prints *at least* information that is enough to identify the last test (which triggered a SEGV) *immediately*. Better if it will print all available information once it is available and have only a summary printed from `handler.suiteEnd`.

tarruda:
Here's one that seems to do the trick:

```lua
return function(options, busted)
  local handler = require 'busted.outputHandlers.utfTerminal'(options, busted)

  handler.testStart = function(element, parent, status, debug)
    io.write('\n' .. handler.getFullName(element) .. ' ... ')
    io.flush()
  end

  busted.subscribe({ 'test', 'start' }, handler.testStart)

  return handler
end
```

I'm gonna add it to neovim deps until there's a native implementation in busted

tarruda:
This variant also outputs the file name before the tests:

```lua
local ansicolors = require 'ansicolors'

return function(options, busted)
  local handler = require 'busted.outputHandlers.utfTerminal'(options, busted)

  handler.fileStart = function(name)
    io.write('\n' .. ansicolors('%{cyan}' .. name) .. ':')
  end

  handler.testStart = function(element, parent, status, debug)
    io.write('\n  ' .. handler.getFullName(element) .. ' ... ')
    io.flush()
  end

  busted.subscribe({ 'file', 'start' }, handler.fileStart)
  busted.subscribe({ 'test', 'start' }, handler.testStart)

  return handler
end
```

thefosk:
+1

tarruda:
This is already available in the latest version with the `gtest` output handler

o-lim:
I believe PR #390 and #394 address this issue.

PR #390, opened 2015-03-19
Update TAP output handler

### LINKED ISSUE #345: Provide more information to testEnd.
I'd like to provide a bit more information than just the stack trace when something fails in the TAP and junit output.  Unfortunately, it seems the nice error message about the failure isn't present until the end of the suite.  Is there any chance this information can be added to the `element` node in the `testEnd` handlers?  I'd do this myself, but I'm still learning lua and busted is quite abstract.

o-lim:
This updates the TAP output handler to output the result of each test as it completes, instead of displaying all test results at the end of a suite. Also update TAP output to report skipped tests and errors occurring outside of tests.
```
$ busted -o TAP
ok 1 - Tests success1
ok 2 - Tests success2
not ok 3 - Tests failure1
# spec/test_spec.lua @ 12
# Failure message: spec/test_spec.lua:13: always fails
not ok 4 - Tests failure2
# spec/test_spec.lua @ 16
# Failure message: spec/test_spec.lua:17: always fails
not ok 5 - Tests error1
# spec/test_spec.lua @ 20
# Failure message: spec/test_spec.lua:21: always throws
not ok 6 - Tests error2
# spec/test_spec.lua @ 24
# Failure message: spec/test_spec.lua:25: always throws
ok 7 - # SKIP Tests pending1
ok 8 - # SKIP spec/test_spec.lua:32: pending2
not ok 9 - Tests teardown
# spec/test_spec.lua @ 2
# Failure message: spec/test_spec.lua:3: error in teardown
1..9
```
```
$ busted -v -o TAP
ok 1 - Tests success1
ok 2 - Tests success2
not ok 3 - Tests failure1
# spec/test_spec.lua @ 12
# Failure message: spec/test_spec.lua:13: always fails
# stack traceback:
#       spec/test_spec.lua:13: in function <spec/test_spec.lua:12>
#
not ok 4 - Tests failure2
# spec/test_spec.lua @ 16
# Failure message: spec/test_spec.lua:17: always fails
# stack traceback:
#       spec/test_spec.lua:17: in function <spec/test_spec.lua:16>
#
not ok 5 - Tests error1
# spec/test_spec.lua @ 20
# Failure message: spec/test_spec.lua:21: always throws
# stack traceback:
#       spec/test_spec.lua:21: in function <spec/test_spec.lua:20>
#
not ok 6 - Tests error2
# spec/test_spec.lua @ 24
# Failure message: spec/test_spec.lua:25: always throws
# stack traceback:
#       spec/test_spec.lua:25: in function <spec/test_spec.lua:24>
#
ok 7 - # SKIP Tests pending1
ok 8 - # SKIP spec/test_spec.lua:32: pending2
not ok 9 - Tests teardown
# spec/test_spec.lua @ 2
# Failure message: spec/test_spec.lua:3: error in teardown
# stack traceback:
#       spec/test_spec.lua:3: in function <spec/test_spec.lua:2>
#
1..9
```

This should fix #345 and helps with #375.

### DIFF
--- a/busted/outputHandlers/TAP.lua
+++ b/busted/outputHandlers/TAP.lua
@@ -3,50 +3,82 @@ local pretty = require 'pl.pretty'
 return function(options, busted)
   local handler = require 'busted.outputHandlers.base'(busted)
 
+  local success = 'ok %u - %s'
+  local failure = 'not ' .. success
+  local skip = 'ok %u - # SKIP %s'
+  local counter = 0
+
+  handler.suiteReset = function()
+    counter = 0
+    return nil, true
+  end
+
   handler.suiteEnd = function()
-    local total = handler.successesCount + handler.errorsCount + handler.failuresCount
-    print('1..' .. total)
+    print('1..' .. counter)
+    return nil, true
+  end
 
-    local success = 'ok %u - %s'
-    local failure = 'not ' .. success
-    local counter = 0
+  local function showFailure(t)
+    local message = t.message
+    local trace = t.trace or {}
 
-    for i,t in pairs(handler.successes) do
-      counter = counter + 1
-      print(success:format(counter, t.name))
+    if message == nil then
+      message = 'Nil error'
+    elseif type(message) ~= 'string' then
+      message = pretty.write(message)
     end
 
-    local showFailure = function(t)
-      counter = counter + 1
-      local message = t.message
-      local trace = t.trace or {}
-
-      if message == nil then
-        message = 'Nil error'
-      elseif type(message) ~= 'string' then
-        message = pretty.write(message)
-      end
-
-      print(failure:format(counter, t.name))
-      print('# ' .. t.element.trace.short_src .. ' @ ' .. t.element.trace.currentline)
-      if t.randomseed then print('# Random seed: ' .. t.randomseed) end
-      print('# Failure message: ' .. message:gsub('\n', '\n# '))
-      if options.verbose and trace.traceback then
-        print('# ' .. trace.traceback:gsub('^\n', '', 1):gsub('\n', '\n# '))
-      end
+    print(failure:format(counter, t.name))
+    print('# ' .. t.element.trace.short_src .. ' @ ' .. t.element.trace.currentline)
+    if t.randomseed then print('# Random seed: ' .. t.randomseed) end
+    print('# Failure message: ' .. message:gsub('\n', '\n# '))
+    if options.verbose and trace.traceback then
+      print('# ' .. trace.traceback:gsub('^\n', '', 1):gsub('\n', '\n# '))
     end
+  end
 
-    for i,t in pairs(handler.errors) do
-      showFailure(t)
-    end
-    for i,t in pairs(handler.failures) do
-      showFailure(t)
+  handler.testStart = function(element, parent)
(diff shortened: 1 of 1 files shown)


tarruda:
:+1: 

I have a suggestion: Print the test name before it is executed and a status string after, something along these lines:

Before: 
```
some test ..
```

After:

```
some test .. {ok/fail/pending/error}
```

A possibility is to refactor utfTerminal/plainTerminal to allow subclasses that override the event handlers. I Neovim I have added [this one](https://github.com/neovim/neovim/blob/master/third-party/utfTerminalDetailed.lua), but it could be improved if it was possible to override the "test end" event handler to print a text string instead of a symbol.

tarruda:
BTW here's [utfTerminalDetailed](https://github.com/tarruda/neovim/blob/improve-functional-test-debuggability/third-party/utfTerminalDetailed.lua) updated to the latest busted version

jszakmeister:
> Before:
> 
> some test ..
> After:
> 
> some test .. {ok/fail/pending/error}

That would be nice, but anything that prints to the screen would break parsing the results, and this isn't compatible with [TAP](https://testanything.org/tap-specification.html).  I think it would be better to do:

```text
# some test
ok 1 - some test
# some other test
ok 2 - some other test # SKIP
# another test
not ok 3 - another test
1..3
```

I'd only print the extra `#` lines in verbose mode too.

tarruda:
> That would be nice, but anything that prints to the screen would break parsing the results, and this isn't compatible with TAP.

True, I wasn't considering compatibility with TAP parsers.

In any case I think it would be nice having something like utfTerminalDetailed in busted, it greatly simplifies finding hangs/crashes in tests

jszakmeister:
> In any case I think it would be nice having something like utfTerminalDetailed in busted, it greatly simplifies finding hangs/crashes in tests

A huge :+1:.  I've been looking at similar things with TAP for the same reasons.

o-lim:
Instead of trying to modify the TAP output to do something it was never really designed to do, I created a another output handler that mirrors the gtest unit test output format in PR #394.

jszakmeister:
> Instead of trying to modify the TAP output to do something it was never really designed to do, I created a another output handler that mirrors the gtest unit test output format in PR #394.

I wouldn't say that.  TAP is quite flexible and I've seen it used in interesting ways.  Plus, it's standard format is understood by many tools (part of my motivation in seeing it improved).  I think the small difference here would be to get this line:

```# spec/test_spec.lua @ 12
```

emitted before running the test so that the user has a clue about what's going on.  I've had something fairly close for a while, but getting a better stack trace on failure has been problematic.

o-lim:
Updated to print test name including file and line number with verbose option.

o-lim:
@jszakmeister: Does this work for you?

jszakmeister:
It looks good to me!  I ran across a bug between QuickBuild and this output, but I think it's QB's fault (TAP support is relatively new in QB).

jszakmeister:
@o-lim Do you need anything else from me?

tarruda:
> @o-lim Do you need anything else from me?

I guess he doesn't have commit access and is just waiting for the PR to be merged.

jszakmeister:
With all the work he has done, I just assumed he did. :-)

Issue #115, opened 2013-04-06
Improved run task loading through .busted


cahna:
I found this beautiful testing framework this evening, but had some issues getting started (mostly with setting up predef'ed tasks), so I haven't even started using it for the project I meant to. Do predefined tasks (pt) outstanding issues?

I noticed the .busted pt file was being loaded and 'eval'ed as Lua code within busted_bootstrap. Since Penlight is being used elsewhere in the project, [I changed this](https://github.com/cahna/busted/commit/323c542e2112aadc46b4f54f4604ebf8d9fe5797#commitcomment-2958229) to a 'safer' import with pl.utils.readfile and pl.pretty.read. 

Not a critical issue given the context, but other formats could easily be included for valid pt in this way (json, ini, ...). I'm also hoping to allow for a 'global' option within the pt file to quickly set options to be inherited by all runs within the file unless explicitly overwritten (ie: './spec' and '_spec.lua' to './test' and '.test.lua'). Is this something you would be interested in?

PR #396, opened 2015-03-26
Improve configuration task loading

o-lim:
This adds a special global option `_all`, to the .busted file, that is inherited by all other configuration options unless explicitly overwritten.
```lua
-- .busted
return {
  _all = {
    coverage = true
  },
  default = {
    verbose = true
  },
  apiUnit = {
    tags = { "api" },
    ROOT = { "spec/unit" },
    coverage = false
  }
}
```
Any command-line flag enabled in the .busted file, except `--version` and `--list`, can be disabled on the command-line with `--no-flags`.  For example using the above .busted file, you can disable `verbose` like so:
```
$ busted --no-verbose
```
Added a command-line option to explicitly specify the configuration file to be loaded.
```
$ busted --config-file=my_config_file.lua
```
or
```
$ busted -f my_config_file.lua
```

Addresses issue #115.


mpeterv:
Why is it called '_all' and not simply 'all'?

o-lim:
It is very possible that 'all' can as the name for a configuration.  For example:
```lua
return {
  all = {
    ROOT = { 'spec' }
  },
  default = {
    ROOT = { 'spec/basic_tests' }
  },
  set1 = {
    ROOT = { 'spec/set1' }
  },
  set2 = {
    ROOT = { 'spec/set2' }
  }
}
```
So I prefixed it with '_' to avoid any possible collisions.

o-lim:
Updated PR to allow disabling of flags via the command-line.

Issue #405, opened 2015-04-09
Please use the HTTPS repo url in the rockspec file


OptiverTimAll:
I'm trying to build NeoVim, which depends on busted. It installs busted by running the following command:

    luarocks build https://raw.githubusercontent.com/Olivine-Labs/busted/master/busted-scm-0.rockspec

The command successfully obtains the `.rockspec` file via our HTTP proxy, but then it tries to clone the listed repo which uses a `git://` URL and therefore fails. If you could update the `.rockspec` file to mention https://github.com/Olivine-Labs/busted.git instead of the `git://` URL, I (and many other people behind restrictive firewalls) would be very appreciative.

o-lim:
Sorry, looks like luarocks does not support https with git.  Couldn't use use the rc rockspec instead?

o-lim:
Why can't you update NeoVim to install busted via `luarocks install busted`?  I believe the `busted-2.0` rockspec uses https, which would solve your problem.  Or you could just manually clone and install busted yourself.
```
git clone https://github.com/Olivine-Labs/busted.git
cd busted
sudo luarocks make busted-scm-0.rockspec
```
You should probably be using the latest released version of busted, rather than cloning the latest master branch.  If there is a feature you need that is not in busted-2.0-rc7, then maybe that means it's about time to release busted-2.0-rc8?

jszakmeister:
> If there is a feature you need that is not in busted-2.0-rc7, then maybe that means it's about time to release busted-2.0-rc8?

Yes, that's exactly the case.  There have been a series of little things that keep us from using the latest released version.  The biggest thing at the moment though is some of the enhanced reporting that has been added.

> Or you could just manually clone and install busted yourself.

We try to keep things easy for contributors, but coordinating something like this would be a real PITA.  OTOH, your current scm spec is broken--it does not do what you intend.

I did get a patch into luarocks (keplerproject/luarocks#350) to fix the issue, but it's probably best to stick to `git+http` for now, as any luarocks that doesn't support `git+https` would fail.

o-lim:
Sounds like the ideal solution might be to get busted-2.0-rc8.  Then you could just use `luarocks install busted`.  In general, I would prefer this over cloning the latest master branch.

I changed the scm rockspec back to use the git protocol since the git+https wasn't supported.

@DorianGray any chances for an rc8 release some time soon?

ajacksified:
rc8 released.

o-lim:
[...]

PR #409, opened 2015-04-14
Update rockspec to use https url for cloning

o-lim:
Fixes issue #405

### DIFF
--- a/busted-scm-0.rockspec
+++ b/busted-scm-0.rockspec
@@ -1,7 +1,7 @@
 package = 'busted'
 version = 'scm-0'
 source = {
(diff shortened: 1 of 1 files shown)


jszakmeister:
Luarocks appears to have trouble with this change as it cannot determine correctly if the url is a download of an archive, git, hg, or something else.  So it treats it like an archive, which is wrong.  It looks like you're meant to do something like `url = "git+https://github.com/Olivine-Labs/busted.git"` instead, but that doesn't work because luarocks doesn't have a git_https fetch module.  However, `url = "git+http://github.com/Olivine-Labs/busted.git"` does appear to work--though it's getting redirected.

PR #416, opened 2015-05-08
Fix specs to match latest luassert changes

o-lim:
This updates specs to pass against the latest changes made to luassert.

Note: The travis build will fail until a new version of luassert is released.


DorianGray:
just waiting on luassert release

PR #422, opened 2015-05-30
Refactor error type detection

o-lim:
This simplifies the error status detection and make it more robust in case a user defined error object has a similar metatable.


o-lim:
Rebased on top of PR #424 so that the travis build would pass.

PR #423, opened 2015-05-30
Wrap publish start/end messages in a safe call

o-lim:
This wraps publishing of `start` and `end` messages in a `safe` call.  By wrapping `publish` in `safe` if an error is thrown in a `start` message handler, the `end` message will still be published and the current block will be skipped.  This makes `start` and `end` messages work similar to `setup` and `teardown` functions.


o-lim:
Rebased on top of PR #424 so that the travis build would pass.

PR #424, opened 2015-05-30
Add luarocks backup server for travis build

o-lim:
This adds a backup server from which to download luarocks, and updates the luarocks version to 2.2.2


o-lim:
Since async tests are not yet supported, removed installation of lib-ev an copas from travis build.  In addition, the latest version of copas does not support Lua 5.3, so the travis build will always fail when trying to install copas 1.2 or later.

PR #427, opened 2015-06-03
Remove ansicolors dependency from rockspec

o-lim:


ajacksified:
Might as well nuke it from .luacov too :D

o-lim:
Removed ansicolors from .luacov too.

o-lim:
Replaced ansicolors with lua-term in .luacov

Issue #238, opened 2014-05-06
First level describe text repeated in nested describe output.


benbarbour:
For example, given

```
describe("Busted bug with describe output", function()
    it("doesn't happen with level0 describes", function()
        assert.is_true(false)
    end)

    describe("nested 1", function()
        it("doesn't happen with level1 describes", function()
            assert.is_true(false)
        end)
        describe("nested 2", function()
            it("happens with level2 describes", function()
                assert.is_true(false)
            end)
        end)
    end)
end)
```

the output is:

```
Busted bug with describe output / nested 1 / Busted bug with describe output / nested 2 / happens with level2 describes
```

PR #434, opened 2015-06-21
Fix repeated values in nested test description output.

Billiam:
This is a fix for the stable branch to address repeated description values in test descriptions. Currently, the deeper a test is nested, the more duplicates appear in each description (until they're unreadable).

This PR adds a test for the output description of specs nested within describe blocks.

Fix changes the `raw_describe` method so that it only adds the current_context to the new context's parent list, rather than also adding all of the current_context's parents as well.

Relates to #238

### DIFF
--- a/spec/core_spec.lua
+++ b/spec/core_spec.lua
@@ -260,6 +260,25 @@ teardown A
 
 end)
 
+it "Nested tests are described correctly" (function()
+  local statuses = busted.run_internal_test(function()
+    describe "Outermost description" (function()
+      describe "Outer description" (function()
+        describe "Inner description" (function()
+          describe "Innermost description" (function()
+            it "Inner assertion" (function()
+              assert(true)
+            end)
+          end)
+        end)
+      end)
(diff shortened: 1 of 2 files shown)


Billiam:
Current test failures: 
* Travis installation of luarocks, see #435 
* `/usr/local/share/lua/5.2/busted/core.lua:67: ./spec/core_spec.lua:9: assertion failed!` There's an assertion (unrelated to this PR) that `mock` is a function, but I believe it's a callable table. This may be due to updates to luassert?

ajacksified:
This is failing because your changes are on a _very_ old version of busted; please rebase, or it might be easier to recreate the new PR based on busted master.

Billiam:
Yep! That is deliberate.

 I was hoping to have this PR merged into the stable branch, specifically, as the 2.0 branch does not have features I currently need (async), so that a new 1.x tag could be released.

I believe this issue is already solved (or is radically different) in master.

PR #438, opened 2015-07-05
Minor refactors

o-lim:
A few refactors
* Default options
* Calls to `os.exit()`
* Use `socket.gettime()` instead of `os.clock()`


DorianGray:
This is solid. Nice work!

PR #439, opened 2015-07-07
Implement async. WIP

DorianGray:


DorianGray:
Currently, only IT blocks have async. More to follow. I need to figure out how to make blocks work properly with disconnected block ends.

DorianGray:
@o-lim @ajacksified , please review.

o-lim:
I'm not sure what you mean by `disconnected block ends`.  Wouldn't you just add `async` to `it`/`setup`/`teardown`/`before_each`/`after_each` blocks?  Wouldn't you just call the event loop inside the block after it returns from `busted.safe(descriptor, element.run, element)`?  `async` just means that the test or block depends on asynchronous inputs, right?  So, it's not like you should start executing the next block/test until the current block/test completes.

Maybe something like:
```lua
    local async, evloop
    element.env.async = function(loop)
      async = true
      evloop = loop
      element.env.done = done.new(loop.stop)
    end
    if pass then
      local status = busted.status('success')
      if busted.safe_publish('it', { 'test', 'start' }, element, parent) then
        status:update(busted.safe('it', element.run, element))
        if status:success() then
          status:update(busted.safe('it', evloop.run, element))
        end
        if finally then
          block.reject('pending', element)
          status:update(busted.safe('finally', finally, element))
        end
      else
        status = busted.status('error')
      end
      busted.safe_publish('it', { 'test', 'end' }, element, parent, tostring(status))
    end
```
Did you want to return `done` instead of injecting it into the environment?  Since `done` is not available until you call `async`, it might make more sense to return `done` in order to avoid possible confusion.

DorianGray:
Yeah, I did want to return done. I'm glad someone else is on board with that, I got some negative feedback when I suggested it earlier. I'm going to take another pass at this this week and see if I can nail this down finally.

ajacksified:
ping @DorianGray :trollface:

adriweb:
Any news about this? :)

DorianGray:
None. I have no time. Someone else please feel free to open a PR.

adriweb:
Aww that's too bad - indeed with the current patch here the tests just seem to go on without waiting on the async one's completion.
Thanks for the quick reply though.

PR #440, opened 2015-07-12
Fix busted globals with strict.lua

nehz:
Get undeclared variable error when using the default strict.lua (unmodified) when environment has not been setup (e.g in setup())

For example

    describe(function()
      setup(function() set_global = 1 end)
      it('test', function() assert.equal(set_global, 1) end)
    end)


DorianGray:
Would you mind cleaning up the commit history and also adding some tests for this behavior?!

PR #441, opened 2015-07-13
Refactor exit code

o-lim:
Always return 1 on error and 0 on success. Avoid calling `os.exit` whenever possible, throwing an `error` to exit with error, and allowing Lua to exit naturally.


DorianGray:
Looks solid. Thanks!

mpeterv:
Ok, so now under Lua 5.3 whenever I want to print help or there is an error I also get this printed after normal output:

```
(error object is a nil value)
stack traceback:
	[C]: in function 'error'
	stdin:1: in main chunk
	[C]: in ?
```


`error()` is not always silent.

PR #450, opened 2015-08-18
Handle errors from __gc metamethods

o-lim:
This handles errors from __gc metamethods as well as any other errors that are not caught by xpcall.

### DIFF
--- a/busted/core.lua
+++ b/busted/core.lua
@@ -173,6 +173,11 @@ return function()
     end) }
 
     if not ret[1] then
+      if status == 'success' then
+        status = 'error'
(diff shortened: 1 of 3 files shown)


DorianGray:
looks nice, will merge as soon as tests run.

o-lim:
Thanks, let me verify that the test runs and passes with Lua 5.2/5.3.  I am only able to test with 5.1 on my system.

o-lim:
Ok, I was able to fix the unit test for Lua 5.2/5.3.  Everything should be good to go now.

DorianGray:
very nice! Thank you!

PR #454, opened 2015-09-20
Added pt-BR language to i18n

sobrinho:


o-lim:
Might also want to add pt-BR to luassert as well.  Otherwise, the assertion failures will be in English, by default.

PR #455, opened 2015-09-24
Refactor calls to module loaders

o-lim:
This just cleans up the calls to the various loaders (i.e. output handler, filter loader, helper loader).


DorianGray:
This looks really good, thanks!

PR #461, opened 2015-10-27
Implement option to choose the lua interpreter

DorianGray:
@ajacksified @leafo @calio @o-lim 

This is a giant hack, but it works! The correct fix is not in busted, but in luarocks. Currently, luarocks wraps binaries in a shell script that forces a particular interpreter on us, the one that luarocks is configured for.

### DIFF
--- a/busted/modules/cli.lua
+++ b/busted/modules/cli.lua
@@ -1,6 +1,8 @@
 local utils = require 'busted.utils'
 local path = require 'pl.path'
 local tablex = require 'pl.tablex'
+local exit = require 'busted.compatibility'.exit
+local execute = require 'busted.compatibility'.execute
 
 return function(options)
   local appName = ''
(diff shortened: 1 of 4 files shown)


leafo:
nice :+1:

o-lim:
`os.execute` returns different arguments for different versions of Lua.  So perhaps using `pl.utils.execute` would be better for compatibility.  Something like:
```lua
local exit = require 'busted.compatibility'.exit
local execute = require 'busted.compatibility'.execute
...
if ... then
  exit(execute(cliArgs['lua']..' '..args[0]..' --ignore-lua '..table.concat(args, ' ')) and 0 or 1)
end
```

busted/modules/cli.lua
```diff
@@ -129,6 +129,7 @@ return function(options)
   cli:add_option('--lang=LANG', 'language for error messages', 'en', processOption)
   cli:add_option('--loaders=NAME', 'test file loaders', defaultLoaders, processLoaders)
   cli:add_option('--helper=PATH', 'A helper script that is run before tests', nil, processOption)
+  cli:add_option('--lua=LUA', 'The full path to the lua interpreter busted should run under', nil, processOption)
```
o-lim: Technically, it doesn't have to be the full path, right?  It should still work if Lua was in the $PATH (i.e. `lua51`, `lua52`, `lua53`)

calio:
Nice, hacky but it works! Agreed with @o-lim, it should also work if Lua interpreter is in the $PATH.

DorianGray:
@o-lim better?

DorianGray:
I removed the or 255. I am seeing some weirdness in return codes still but we'll need to fix that in another PR.

PR #462, opened 2015-10-27
Ensure that pre-release versions of lua_cliargs 3 aren't used as a dependency

ryanplusplus:
lua_cliargs 3.0.rc-1 was released a few hours ago and unfortunately Luarocks determined that this was < 3.0. So, if you install the current busted 2.0 RC it will install this incompatible version of lua_cliargs.

This change ensures that no 3.0 RCs are installed for lua_cliargs.

### DIFF
--- a/busted-2.0.rc10-1.rockspec
+++ b/busted-2.0.rc10-1.rockspec
@@ -19,7 +19,7 @@ description = {
 }
 dependencies = {
   'lua >= 5.1',
-  'lua_cliargs >= 2.5-0, < 3.0',
+  'lua_cliargs >= 2.5-0, < 3.0.rc-1',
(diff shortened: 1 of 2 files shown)


amireh:
That would explain the errors i'm seeing on travis for the 2.5 branch of cliargs. It's a funny dependency party going on here... https://travis-ci.org/amireh/lua_cliargs/jobs/87703027

o-lim:
This still doesn't work.  Maybe you need to do `< 3` instead?

ryanplusplus:
I think this should be fixed now. The issue was that I only made the change in the release rockspec, not the SCM rockspec.

DorianGray:
Thanks!

PR #480, opened 2016-01-22
Use a Lua C module for time related functions

### LINKED ISSUE #472: Remove luasocket dependency
I'm trying to use busted to test code that does different things depending on whether luasocket is installed or not. The luasocket dependency is quite large; and only used for `socket.gettime()` (which returns the current time with microsecond precision)

o-lim:
This removes the dependency on luasocket by adding a C function for `gettime`, instead of using `socket.gettime`. Basically, extracting the time related functions from luasocket and using them to create a Lua C module for busted.

This should resolve #472.


Tieske:
bad idea... introducing c code into busted.

If using this, then it should be an external module. And then, it would probably be better to think what other very common elements are missing in the lua eco-system and build one lib for that. (eg. randomization, time, copy files iirc)

DorianGray:
I agree, we should definitely pull this out into a separate module. We also need to verify that it works on arm, and x86 32 + 64bit with optimally mips included also.

o-lim:
Sounds good to me.  What would you want to call it and where would you put it?  The code was basically copied from luasocket, so it should work everywhere luasocket works.

DorianGray:
cool, we can drop it in the OL repo if you'd like, or keep it on your local if you want to. It just needs to be available on luarocks.

DorianGray:
You should probably have push access anyways

o-lim:
Dropping it into the OL repo sounds good to me.  It's a really small module, so it shouldn't need much maintenance.

o-lim:
I created a repo for the `gettime` function https://github.com/o-lim/luasystem.  It just provides `gettime` and `sleep` functions as a Lua C module.  I didn't know what to call it so I just named it `luasystem`.  If you don't like the name I can rename it to something else.  And if you guys are okay with everything in it, I can make an initial release and push it to luarocks, then update this PR to use it.

DorianGray:
Very nice! It looks great and is exactly what we need.

o-lim:
Closing this PR and replacing with #488, which uses luasystem to replace luasocket.

daurnimator:
Hmm. I actually would have preferred this going into busted.
Now we just have another C module >.<

xpol:
I'd really glad to see luasocket be removed. It takes too much time to install in ci infrastructure like travis.

Issue #473, opened 2016-01-13
refactor timing code into the core


Tieske:
Currently the timing is implemented 3 times in different files (output handlers)

The code should be refactored such that the time measurements are in the core, and the results are exposed to the output handlers.

Tieske:
fixed by #483

PR #483, opened 2016-01-26
Refactor test execution timing

o-lim:
Move the computation and recording of test execution time to the core and insert the start/end time and duration of each block into its context for output handlers to consume and report.

This should resolve issue #473

### DIFF
--- a/busted/outputHandlers/junit.lua
+++ b/busted/outputHandlers/junit.lua
@@ -15,16 +15,14 @@ return function(options)
   }
   local stack = {}
   local testcase_node
-  local testStartTime
-  local output_file_name
   if 'table' == type(options.arguments) then
     --the first argument should be the name of the xml file.
     output_file_name = options.arguments[1]
   end
 
   handler.suiteStart = function(suite, count, total)
-    local suite = {
-      start_time = busted.gettime(),
+    local suite_xml = {
+      start_time = suite.starttime,
       xml_doc = xml.new('testsuite', {
         name = 'Run ' .. count .. ' of ' .. total,
         tests = 0,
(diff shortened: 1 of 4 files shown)


busted/outputHandlers/junit.lua
```diff
@@ -37,13 +35,17 @@ return function(options)
...
+
   local function elapsed(start_time)
-    return string.format("%.2f", (socket.gettime() - start_time))
+    return formatDuration(busted.gettime() - start_time)
   end
 
   handler.suiteEnd = function(suite, count, total)
     local suite = top
-    suite.xml_doc.attr.time = elapsed(suite.start_time)
+    suite.xml_doc.attr.time = formatDuration(suite.duration)
```
dantheho: Line 47 and 48 need to be swapped, the 'top' data structure does not have suite.duration.

o-lim:
Fixed the error in the junit output handler.

o-lim:
Updated to resolve merge conflicts

DorianGray:
I would like this to not be in the "core", but instead be entirely a function of the output handlers. That way, we could allow output handlers to skip timing code if they don't use it...not to mention we'd shrink the core a bit.

Is this realistic? Or do you think it would introduce too many complications?

o-lim:
It is kind of nice to have the timing information already there, especially if all/most of the output handlers are using the same information, making it easier to maintain across multiple output handlers instead of having the same code duplicated in each one.  However, I do see your point in that not all output handlers may want to use that information or some may want to compute it differently using a different clock source.

Maybe one way to look at it is that the core is providing information about the test(s), while the output handlers are just formatting that information for display.  If the output handlers perform all of the timing, then they are also generating and injecting that information into the output rather than just reformatting the data.  I think, ideally, the output handlers should just reformat the data, and not inject any new information.

DorianGray:
Fair enough =)

Issue #451, opened 2015-08-21
Add a command line option for writing junit output to a file.


jszakmeister:
This doesn't affect me anymore since I've switched Neovim's QB instance to use TAP, but I figured I'd suggest it anyways.  One of the issues with simply catching stdout and redirecting it to a file is that the underlying app may write data to stdout.  When it does, it corrupts the xml file and most tools just barf on it.  It would be better to provide an option to have the junit handler write the results directly to a file to avoid the intermixing of outputs and keep the XML well-formed.

jimmyharris:
This feature would greatly help me out. I am trying to integrate busted's junit output to an internal tool that knows how to work with junit and not tap. Piping this output to stdout is not an option for the reasons specified above.  I have a simple patch that will write the xml output of the junit plugin to the file name specified in the first `-Xoutput` argument. I will submit a pull request but I don't believe anyone has attempted to cover testing for this tool.

justinmk:
@jszakmeister I believe this was closed by https://github.com/Olivine-Labs/busted/pull/485. 

Seems like QB sometimes has issues with TAP, maybe junit will help...

jszakmeister:
> Seems like QB sometimes has issues with TAP, maybe junit will help...

Actually, it's not QB that has problems...  when the tests segfault and die then the final TAP report information is not presented, and that's when QB has trouble.  I'm not sure junit will help here, because the XML will be malformed in the event a segfault happens too. :-(

PR #484, opened 2016-01-28
Junit file output

jimmyharris:
This is a very simple patch to allow the junit output handler to write its output to a named file rather than stdout. This is a simple method of addressing #451 but I am open to suggestions on how to improve this. It would be convenient if, instead of using `-Xoutput` we had a dedicated top level cli argument for specifying where to save the file but this works for me for the moment.


jimmyharris:
crap. I will re-do this. I think I have something weird going on with my commit history.

PR #485, opened 2016-01-28
Allow junit output handler to save output to file

jimmyharris:
When trying to hook busted's junit output module up to an internal
automated test system it is very challenging to determine what output
was written to stdout by tests, and what output was written to stdout by
tests and what output is part of the test summary in xml. The simplest
way to handle this is to allow junit to accept an argument naming a file
where the xml output should be written via the -Xoutput mechanism. This
patch adds this change to junit.

### DIFF
--- a/busted/outputHandlers/junit.lua
+++ b/busted/outputHandlers/junit.lua
@@ -17,6 +17,11 @@ return function(options)
   local stack = {}
   local testcase_node
   local testStartTime
+  local output_file_name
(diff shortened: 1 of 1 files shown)


jimmyharris:
This should address #451. Like I said in accidental #484. I am open to feedback on how to improve this functionality. currently it would work like this

```
busted --output=junit -Xoutput outputFile.xml
```

Then the file would be written to outputFile.xml.

PR #486, opened 2016-02-02
Fix to junit timestamp string format:

cjtallman:
I'm testing version 2.0.rc11-0 and having an issue with the junit output handler crashing lua.
After a little digging, I found it's related to the call to os.date to format the junit timestamp.  The date format string is `'!%Y-%m-%dT%T'`, however the %T format specifier doesn't seem to be available on my platform. Changing the %T to %H:%M:%S in accordance to the strftime documentation resolves the issue.

### DIFF
--- a/busted/outputHandlers/junit.lua
+++ b/busted/outputHandlers/junit.lua
@@ -27,7 +27,7 @@ return function(options)
         errors = 0,
         failures = 0,
         skip = 0,
-        timestamp = os.date('!%Y-%m-%dT%T'),
+        timestamp = os.date('!%Y-%m-%dT%H:%M:%S'),
       })
     }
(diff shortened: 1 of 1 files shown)


ajacksified:
Seems reasonable; as an aside, what OS and version of lua are you using?

cjtallman:
Sure, my current configuration is this:

OS: Windows 7 x86_64
Compiler: MSVC 2010
Lua: v 5.1.5
Luarocks: v 2.3.0

cjtallman:
Thanks mate.

I guess my comment that %T doesn't work on my _platform_ is not technically correct.  It's actually the MSVC 2010 runtime that doesn't support the %T format.  My other configuration using MSVC 2015 does support %T.

PR #488, opened 2016-02-11
Remove luasocket depenency

### LINKED ISSUE #472: Remove luasocket dependency

o-lim:
This removes the dependency on luasocket by using luasystem for `gettime`, instead of using `socket.gettime`.

This replaces PR #480, and should resolve #472.


daurnimator:
Do *not* use gettime. it's causes problems if you use NTP or any other clock programs.
As I mentioned in #472, you need to use a monotonic time source (and I even linked to code!)

xpol:
I'd really glad to see luasocket be removed. It takes too much time to install in ci infrastructure like travis.

Issue #490, opened 2016-02-15
require 'busted.runner'() fails


romanchyla:
As said, this will fail with the foloowing error:

```
lua tests/iotzfutils_spec.lua 
lua: ...al/share/lua/5.1/busted/modules/test_file_loader.lua:25: bad argument #1 to 'find' (string expected, got nil)
stack traceback:
	[C]: in function 'find'
	...al/share/lua/5.1/busted/modules/test_file_loader.lua:25: in function 'pred'
	/usr/local/share/lua/5.1/pl/tablex.lua:634: in function 'filter'
	...al/share/lua/5.1/busted/modules/test_file_loader.lua:24: in function 'getTestFiles'
	...al/share/lua/5.1/busted/modules/test_file_loader.lua:45: in function 'getAllTestFiles'
	...al/share/lua/5.1/busted/modules/test_file_loader.lua:60: in function 'testFileLoader'
	/usr/local/share/lua/5.1/busted/runner.lua:167: in function </usr/local/share/lua/5.1/busted/runner.lua:9>
	tests/iotzfutils_spec.lua:1: in main chunk
	[C]: ?
```

the error happens because the pattern is 'nil', if I specify patter in my .busted config, it works

(when i run the code using 'busted' it works fine without the require) - I think the getTestFiles() funciton should detect/deal with the pattern being 'nil'

my version is:  2.0.rc10-0 (installed) - /usr/local/lib/luarocks/rocks

o-lim:
I couldn't reproduce the same error, but just in case, this should be fixed with PR#493.  You might want to try to see if this also happens with rc11.

romanchyla:
Wow, thanks! I will try to test and get back if there is some new development.

Issue #491, opened 2016-02-15
debugging inside Eclipse almost works


romanchyla:
Using Eclipse LDT is possible, however I found two issues:

- when coverage is enabled, the debugger doesn't stop on breakpoint [ok, no big deal ]
- however, with the remote debugger, the busted.runner fails at around line 10

```
options.defaultOutput = term.isatty(io.stdout) and 'utfTerminal' or 'plainTerminal'
```

it fails because 'io.stdout' is not a FILE but a table [with the remote debugger tcp connection]

a possilbe solution:

```
options.defaultOutput = type(io.stdout) == 'file' and term.isatty(io.stdout) and 'utfTerminal' or 'plainTerminal'
```

PR #493, opened 2016-02-15
Fixes for using debugger

o-lim:
This fixes errors that occur when running with a Lua debugger.

Note: A full fix will also require an update to luassert, which fixes the same type of error (see Olivine-Labs/luassert#132).

This should fix issues #490 and #491.


romanchyla:
thanks a lot!

@o-lim maybe enhance the documentation too? (saying that for the debugger to work, the coverage has to be deactivated?)

DorianGray:
We also need to update the docs for this... @ajacksified ?

PR #494, opened 2016-02-19
Output error message for bad configuration files

o-lim:
Check that any configuration file that was specified is loaded successfully, instead of failing silently and ignoring the file. When there is an error loading in the configuration file, output an error message and exit.


DorianGray:
Very nice!

PR #503, opened 2016-03-11
Use new travis infra as per moteus/lua-travis-example

istr:
Based on https://github.com/moteus/lua-travis-example this PR changes the travis setup and drops ```sudo``` in favor of the faster container-based travis infrastructure. Adds checks for luajit and luajit 2.1.


DorianGray:
Thanks!

PR #508, opened 2016-04-09
Improve randomness by using /dev/urandom

o-lim:
This also ensures that a different random seed is used when repeat is greater than 1 and a seed is not specified. If `/dev/urandom` is not available, then default to `os.time()`.


DorianGray:
This looks good!

PR #540, opened 2017-01-01
Make busted capitalization more consistent

CrackedP0t:
It appears that "busted" is normally spelled with a lowercase 'b'. It was capitalized in the header.


ajacksified:
Thanks!

PR #554, opened 2017-07-01
TAP handler: flush output immediately

justinmk:
8a4d223c5ae4ace8d88d4ee3ca79020f6e67ae21 changed the TAP handler to
write results after each test, but the output is still buffered.

If a test hangs, the buffered results of many previous tests will never
be printed, and it's not clear which test caused the hang. Flushing the
output avoids this.

cc @o-lim

### DIFF
--- a/busted/outputHandlers/TAP.lua
+++ b/busted/outputHandlers/TAP.lua
@@ -16,6 +16,7 @@ return function(options)
 
   handler.suiteEnd = function()
     print('1..' .. counter)
+    io.flush()
     return nil, true
   end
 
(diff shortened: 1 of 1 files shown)


o-lim:
👍

justinmk:
FWIW we have been using this approach without issue in the neovim test suite. Flushing to the output device won't be the bottleneck unless one has hundreds of thousands of extremely fast test cases. The primary purpose of TAP is visibility, debugging hangs is an important factor there.

@ajacksified @DorianGray

DorianGray:
Thanks! Looks good.

PR #561, opened 2017-11-03
Update private functions and CLI usage docs

bonidjukic:
Requiring `mymodule` at the top level executes the code from the module at
that instance, so when later we define `_G._TEST = true` within the `setup`
function, `mymodule` has already been required and the global `_TEST` is
not set.

Solution is to declare the `mymodule` at the top, and then require it within
the `setup` function once the global `_TEST` has been set.


Tieske:
good catch!

DorianGray:
Thank you!

PR #573, opened 2018-06-12
replace luasystem with lua-posix

karlp:
luasystem is unmaintained, doesn't cross compile properly, and only exposes
three functions.  All of which can be implemented using lua-posix, which
is well maintained and portable.

Signed-off-by: Karl Palsson <karlp@etactica.com>

Note: I've not udpated the rockspec files as I don't know how you want to change dependencies there with versions, if at all

### DIFF
--- a/busted/core.lua
+++ b/busted/core.lua
@@ -3,7 +3,8 @@ local setfenv = require 'busted.compatibility'.setfenv
 local unpack = require 'busted.compatibility'.unpack
 local path = require 'pl.path'
 local pretty = require 'pl.pretty'
-local system = require 'system'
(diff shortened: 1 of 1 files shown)


Tieske:
@karlp Does this work on Windows? afaik lua-posix doesn't work on Windows

Tieske:
then that's a no-go...

karlp:
No problem, I'll still patch it out for busted on openwrt, luasystem's build system is useless for cross compiling, and it's an easy patch :)  (I'd recommend just bringing an equivalent implementation inside busted honestly, but I see you've used this as a way of having no C code in _this_ project ;)

PR #587, opened 2019-02-03
Fix typo in cli option processing for config-file

notCalle:
There seems to be a typo in the callback name for `--config-file` option processing.


Tieske:
@ajacksified @DorianGray can we merge this? this is actually a bug.

Issue #150, opened 2013-05-22
src/core.lua should filter out spec files from hidden dirs (like .svn)


lipp:
the penlight dir.getallfiles returns ALL files. hidden files and files from hidden folders should not be executed.
If tests are under VC, the old revisions are executed too, e.g.

./spec/.svn/text-base/foo_spec.lua.svn-base

PR #594, opened 2019-06-02
Allow ROOT to contain hidden directories.

LiamLeFey:
Issue #150 added the functionality of filtering out hidden directories by default.
Sometimes your test cases are in a hidden directory.
This allows you to explicitly add hidden directories as part of ROOT, while still filtering out hidden sub-directories.


Tieske:
looks good, but an additional test to validate the behaviour is missing.

Tieske:
LGTM

PR #601, opened 2019-06-28
fix source URL in rockspec for v2.0.0 release

zackse:
As of the new release, the version referenced in the rockspec file's source URL is inaccessible. This commit changes it to match the tarball version on github (no `-0` suffix). I'm not sure if this is preferable to renaming the archive and its contained directory, but it works in local testing.

### DIFF
--- a/rockspecs/busted-2.0.0-0.rockspec
+++ b/rockspecs/busted-2.0.0-0.rockspec
@@ -1,8 +1,8 @@
 package = 'busted'
 version = '2.0.0-0'
 source = {
-  url = 'https://github.com/Olivine-Labs/busted/archive/v2.0.0-0.tar.gz',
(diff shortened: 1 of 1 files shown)


Tieske:
the tags in the git-repo should NOT contian the LuaRocks rockspec revision number. So this change is valid.

But to get this done, the rockspec should be renamed with an incremented LR revision number; `-1` in this case. And then that rockspec should be pushed/published to the LuaRocks server.

Tieske:
(the quick and dirty fix would be to create another tag `2.0.0-0` (on the same commit as currently tagged `2.0.0`)

Tieske:
replaced by #603

Tieske:
this can be closed now since #603 was merged

zackse:
Thanks for the quick fix in #603.

PR #603, opened 2019-07-01
fix(rockspecs) proper usage of rockspec revision numbers

Tieske:
fixes #602 

After merge, the new rockspecs and rock (rock only for `2.0.0-1`, not for `scm-1`) should be uploaded to LuaRocks

PR #604, opened 2019-07-01
fix(ci) do not specify a rockspec, just build the only one there

Tieske:
Since the published rockspecs are in their own dir, the SCM one is the only one in the main folder and hence a `luarocks make` is enough to build the currently checked out branch.


Tieske:
@ajacksified  This fixes the issue left behind from #603

PR #621, opened 2020-01-14
fix(ci) fix LuaJIT2.0.x setup

Tieske:
the 2.0 version is not installed from source, only the 2.1 was.
despite that the "luajit" option also installs 2.1 (but from a
tar ball)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,7 @@ env:
     - LUA=lua5.2
     - LUA=lua5.3
     - LUA=luajit
-    - LUA=luajit2.0
+    #- LUA=luajit2.0
     - LUA=luajit2.1
 
 before_install:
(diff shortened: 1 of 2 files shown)


Tieske:
I can't figure out what is wrong. Only the build for `luajit2.0` fails, and it fails because somehow a recursive symlink is being generated.
See https://travis-ci.org/Olivine-Labs/busted/jobs/636827962#L302

That output is generated by this line: https://github.com/Olivine-Labs/busted/pull/621/files#diff-d1a673996655db24082d79c375a9c6ecR64

It does not happen with the `2.1.x` `master` branch, nor with the released package for `2.0.5`, but only on the `master` branch for the `2.0.x` version.

Since that version is already released most people will use the released version, and not the master branch, so it should be safe to comment this erroring build out for now.
As was done here: https://github.com/Olivine-Labs/busted/pull/621/files#diff-354f30a63fb0907d4ad57269548329e3R11

So imo should be safe to merge.

@ajacksified @DorianGray thoughts?

PR #631, opened 2020-05-28
fix(luajit) add 'metatype' as a patched method

### LINKED ISSUE #630: Using busted to test something using an FFI lib produces a "cannot change a protected metatable" error when run with luajit
I'm using an FFI library called brinevector (https://github.com/novemberisms/brinevector). When writing unit tests against code using that library, I get a "cannot change a protected metatable" error when running my tests.

It sure looks like it's because busted loads the module twice, but I'm not sure what's going on. Any insight there? Is this a bug in brinevector or in busted?

Tieske:
Abstracts the patching into its own module as well

fixes #630

### DIFF
--- a/busted-scm-2.rockspec
+++ b/busted-scm-2.rockspec
@@ -1,5 +1,5 @@
 package = 'busted'
-version = 'scm-1'
+version = 'scm-2'
 source = {
   url = "git://github.com/Olivine-Labs/busted",
   branch = "master"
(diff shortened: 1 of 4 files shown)


Tieske:
@ajacksified imo this is good to go. But merging it also requires uploading the new SCM-2 rockspec to LuaRocks. Can you merge and upload?

Tieske:
ping @ajacksified

Tieske:
@ajacksified @DorianGray ping again. I do not mind uploading the rockspecs if you can provide me with a token

luarocks.org > login > settings > API keys > generate new key

Issue #632, opened 2020-06-23
TAP output invalid if print function overridden


Luca-spopo:
In my modules, I override the _G.print function to contain extra information such as line numbers. I imagine this is a common scenario. The insulate blocks correctly reset the print function after a test is complete, but any TAP output produced during the test becomes invalid as it uses the overriden function. (This issue may also be present in other output handlers)

The fix is very simple. We can cache the value of the print function in TAP.lua before any test is executed.

Luca-spopo:
Example output:
We can see here that test points 24 to 27 are fine, but then 28 to 30 have test cases which set _G.print to a custom function, which invalidates the output
Invoked using `-v --defer-print -o tap`
`ok 24 - lib.baton.lua Sanity check
# ../tests/baton_spec.lua @ 136: lib.baton.lua active device check - note: Baton does not currently change active device for just mouse inputs
ok 25 - lib.baton.lua active device check - note: Baton does not currently change active device for just mouse inputs
# ../tests/baton_spec.lua @ 230: lib.baton.lua supports dummy table
ok 26 - lib.baton.lua supports dummy table
# ../tests/baton_spec.lua @ 178: lib.baton.lua released and pressed callbacks
ok 27 - lib.baton.lua released and pressed callbacks
Tue Jun 23 23:00:56 2020 ../rocks/share/lua/5.1/busted/outputHandlers/tap.lua 49 : "# ../tests/root_spec.lua @ 14: Exposing fakelove Checking conf.lua is valid"
Tue Jun 23 23:00:56 2020 ../rocks/share/lua/5.1/busted/outputHandlers/tap.lua 60 : "ok 28 - Exposing fakelove Checking conf.lua is valid"
Tue Jun 23 23:00:56 2020 ../rocks/share/lua/5.1/busted/outputHandlers/tap.lua 49 : "# ../tests/root_spec.lua @ 8: Exposing fakelove fakelove raises an error if we print an error"
Tue Jun 23 23:00:56 2020 ../tests/root_spec.lua 9 : "there was an ERROR"
Tue Jun 23 23:00:56 2020 ../rocks/share/lua/5.1/busted/outputHandlers/tap.lua 35 : "not ok 29 - Exposing fakelove fakelove raises an error if we print an error"
Tue Jun 23 23:00:56 2020 ../rocks/share/lua/5.1/busted/outputHandlers/tap.lua 36 : "# ../tests/root_spec.lua @ 8"
[...]

PR #633, opened 2020-06-23
Save print function in TAP output handler before tests

Luca-spopo:
I made this PR to fix https://github.com/Olivine-Labs/busted/issues/632


Tieske:
have you checked other output handlers for the same flaw?

Tieske:
can you update the PR to use io.write?

PR #637, opened 2020-09-22
fix(outputter) cache all functions, use io.write not print

### LINKED ISSUE #632: TAP output invalid if print function overridden

Tieske:
fixes #632
closes #633

### DIFF
--- a/busted/outputHandlers/TAP.lua
+++ b/busted/outputHandlers/TAP.lua
@@ -1,5 +1,14 @@
 local pretty = require 'pl.pretty'
 local io = io
+local type = type
+local string_format = string.format
+local string_gsub = string.gsub
+local io_write = io.write
+local io_flush = io.flush
+
+local function print(msg)
+  io_write(msg .. '\n')
+end
 
 return function(options)
   local busted = require 'busted'
@@ -17,7 +26,7 @@ return function(options)
 
   handler.suiteEnd = function()
     print('1..' .. counter)
-    io.flush()
+    io_flush()
     return nil, true
   end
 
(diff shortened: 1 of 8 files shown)


Tieske:
since this is less well tested this could really use a review.

Anyone interested? @ajacksified @DorianGray @kikito @alerque

busted/outputHandlers/TAP.lua
```diff
@@ -31,12 +40,14 @@ return function(options)
       message = pretty.write(message)
     end
 
-    print(failure:format(counter, t.name))
+    print(string_format(failure, counter, t.name))
```
alerque, line 43: I get what you are doing with al these function replacements, but as I understand the original issue report and PR, the `print()` function itself also needs to be cached, but I don't see that happening in this PR. Am I missing something?

Tieske: more broadly; the outputters should cache the functions they use, since I replaced all usage of `print`, it no longer needs to be cached, since the same applies to other functions it does cache `format`, `gsub`, etc.

alerque: Neverbmind, I mussed L9 entirely! That answers my concern.

busted/outputHandlers/TAP.lua
```diff
@@ -1,5 +1,14 @@
 local pretty = require 'pl.pretty'
 local io = io
+local type = type
+local string_format = string.format
+local string_gsub = string.gsub
+local io_write = io.write
+local io_flush = io.flush
+
+local function print(msg)
```
alerque, line 9: It's kind of confusing that methods are cached one way and used with local names (`string.format()`→`string_format()`) while global functions are not (`print()`→`print()`). Maybe this should be prefixed as a local function `_print()` and used that way throughout? (also `_type` etc.)

alerque (changes requested):

busted/outputHandlers/TAP.lua
alerque: If you don't think this is a good idea I'll be fine with that too. It wouldn't be my coding style preference but it gets the job done. The rest of the PR looks okay so if you decide to dismiss this rename suggestion I'll just approve instead.

Tieske: I get the point but never seen anything prefixing them with _ ? so I'd rather dismiss this one indeed.

alerque: I see them all the time … but maybe it's mostly in my own code. And it might be a hold-over from Haskell and Rust where it is a compiler-enforced naming scheme for shadowed scope or unused values. Also note `luacheck` also processes `_` as a way to discard a value so there is some precedent for scope delineation. And since you can't use primes like in Haskell...

In any case if you don't like it no biggy.

alerque (approved):

busted/outputHandlers/TAP.lua
Tieske: the single "_" to ignore something is a very common Lua pattern

PR #646, opened 2020-11-24
feat(fixtures) add function to locate relative fixtures

Tieske:
By providing a relative path from the current test file it will
return an absolute path to the fixture.

Closes #644

PR #652, opened 2020-11-30
fix(readme): Correct link to test runner badge

### LINKED ISSUE #651: GitHub Actions failed for master branch
After merger of #645 everything should have been green, but since I could only test on my fork rather than this repo it was a bit hard to test. Now with the GH Actions workflow at least existing in the repo it can be run locally and PRs should reflect the status. I'll look into what I missed—probably somewhere I left in a path to my fork or something dumb like that.

Feel free to assign this to me in case I don't get to it tonight and forget about it tomorrow ;-)

alerque:
Fixes #651.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@ busted
 ======
 
 [![Luacheck Lint Status](https://img.shields.io/github/workflow/status/Olivine-Labs/busted/Luacheck?label=Luacheck&logo=Lua)](https://github.com/Olivine-Labs/busted/actions?workflow=Luacheck)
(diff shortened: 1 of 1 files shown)


alerque:
The `luacheck` test that is failing is a result of #646. I looked into fixing it but wasn't exactly sure what should happen. It looks like at least some of the lint warnings are legitimate, but since I didn't trace all the scoping involved or the expected behavior I wasn't quite sure how to fix them. Maybe @Tieske can address that in a separate PR.

PR #660, opened 2021-02-26
Add support for luacov config path

### LINKED ISSUE #659: Support configuring path to luacov config file
There should be a command-line option and a `.busted` config field to set the path to the luacov config.

As this is currently unsupported, luacov is always invoked with no options and defaulting to the `./.luacov` config file or the default config.

nicholaschiasson:
Closes #659 

Adding support for a cli or config option `coverage-config-file` to allow setting the path the your luacov config file, in case it is not at `.luacov` in the project root.


DorianGray:
LGTM!

alerque (approved):

nicholaschiasson:
@DorianGray If I may ask, what is the release cycle for this project and when could we expect this merged?

DorianGray:
as soon as someone else takes a look. We're ready to make a release now...we were waiting on a fix from someone but they seem to have disappeared

DorianGray:
we will do a release soon. Apologies for not merging - new baby 4 days ago, no time to test

nicholaschiasson:
Okay no problem! That sounds good. And congratulations on the new baby! 🎂

Tieske:
looks good. One question; if a config-file is provided and the Luacov option is not enabled, what should happen?

- an error because your specifying an unused config file
- automatically enable luacov
- nothing, just execute without luacov

nicholaschiasson:
> if a config-file is provided and the Luacov option is not enabled, what should happen?

I am of the mind that it should have no effect and just execute without luacov, and not just because it is the easiest implementation.

I think it makes sense from a workflow perspective to be able to have the path to your coverage config file configured but still run the framework without running coverage. Say you clone of a project without an adequate development environment (no Luacov installed), it's easier to just run without `--coverage`, or with `--no-coverage`, or remove `coverage = true` from the busted config file, than it is to remove the setting for coverage config file. That's my opinion at least.

Tieske:
@nicholaschiasson that seems fair.

Can you add a test? I think if you add a test to the `cl_spec.lua` file that runs a basic test with a luacov config file specified with a custom output name, and then verify that the output files are actually generated, then that should be good enough.

nicholaschiasson:
@Tieske It looks like the github actions environment is not able to create files. Maybe there's still something I'm doing wrong. Do you have any suggestions?

Tieske:
@nicholaschiasson instead of in the local directory store it as a temp file?

Tieske:
am I right that the problem wasn't creating the files, but that LuaCov wasn't installed?

nicholaschiasson:
> am I right that the problem wasn't creating the files, but that LuaCov wasn't installed?

That's right that was it :sweat_smile: Took me far too long to realize!

Tieske (approved):

Tieske:
Thx @nicholaschiasson !

PR #662, opened 2021-03-25
feat(fixtures) add helpers to read/load fixtures

Tieske:
Also moves the `fixture_path` function (see #646) from a global to the 'fixtures' module. This is non-breaking since that hasn't been released yet.

I really don't like the global `fixture_path` hence the move into a separate module

```lua
local fixtures = "busted.fixtures"

local absolute_path = fixtures.path("../file/some/where.txt")
local my_file = fixtures.read("../file/some/where.txt")
local my_mod = fixtures.load("../some/path/file.lua")
```

### DIFF
--- a/busted/fixtures.lua
+++ b/busted/fixtures.lua
@@ -71,7 +71,7 @@ function fixtures.load(rel_path)
     error(("Error loading file '%s': %s"):format(tostring(rel_path), tostring(err)), 2)
   end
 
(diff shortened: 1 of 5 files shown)


Tieske:
@DorianGray @ajacksified wdyt, can we merge this?

DorianGray:
I definitely like this 1000% better

PR #667, opened 2021-09-05
Issue 666: restore globals set to nil in insulate block

### LINKED ISSUE #666: Insulation does not recover globals set to nil
Noticed this while testing the behaviour of my code when running in a sandbox (i.e. checking if `load`/`loadstring`/`io` are available): busted will delete globals set in an insulated block, but it won't recover values that are set to nil in a block. I've created a minimal, reproducible example to verify this behaviour:

```lua
describe("insulation", function()
    insulate("when value is set to nil", function()
        it("is there in the first place", function()
            assert.is_not_nil(io)
        end)

        it("is set to nil", function()
            _G.io = nil
            assert.is_nil(io)
        end)
    end)

    it("is restoring values afterwards", function() 
        assert.is_not_nil(io)
    end)
end)
```

According to the documentation, 

> Test environment insulation saves the global table _G and any currently loaded packages package.loaded, restoring them to their original state at the completion of the insulate block.

so I would expect `io` to be reset to its original value (the io library) after the `insulate` block, so that all tests should succeed. They don't, however:

```
$ busted mre.lua
●●◼
2 successes / 1 failure / 0 errors / 0 pending : 0.002791 seconds

Failure → mre.lua @ 13
insulation is restoring values afterwards
mre.lua:14: Expected objects to not be the same.
Passed in:
(nil)
Did not expect:
type nil
```

I feel that this issue is similar to #534 in nature, but still distinct.

Henkoglobin:
Seems to be an easy fix, though I'd love feedback on [the `if` in line 22](https://github.com/Olivine-Labs/busted/pull/667/files#diff-5334561e9923d607069a41208ac1b79280fa64d315d6335ae9561ea3809dde60R22) of context.lua: I _don't think_ that there's a reason to restore if `rawget(_G, k) ~= nil` (as then it'd be restored in the first loop), right? 

Fixes #666

### DIFF
--- a/spec/insulate-expose_spec.lua
+++ b/spec/insulate-expose_spec.lua
@@ -35,6 +35,8 @@ describe('Tests insulation', function()
 end)
 
 insulate('Tests insulation with values set to nil', function()
+  -- The tests in this block need to be run in order, as we're testing state recovery between tests.
+
   _G.some_global = true
 
   describe('environment before insulate', function()
--- a/spec/insulate-expose_spec.lua
+++ b/spec/insulate-expose_spec.lua
@@ -34,6 +34,32 @@ describe('Tests insulation', function()
   end)
 end)
 
+insulate('Tests insulation with values set to nil', function()
+  _G.some_global = true
+
+  describe('environment before insulate', function()
+    it('has global', function()
+      assert.is_not_nil(some_global)
+      assert.is_not_nil(_G.some_global)
+    end)
+  end)
+
+  insulate('environment inside insulate', function()
+    it('sets global to nil', function()
+      _G.some_global = nil
+      assert.is_nil(some_global)
+      assert.is_nil(_G.some_global)
+    end)
+  end)
+
+  describe('environment after insulate', function()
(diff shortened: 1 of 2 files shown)


spec/insulate-expose_spec.lua
```diff
@@ -34,6 +34,32 @@ describe('Tests insulation', function()
   end)
 end)
 
+insulate('Tests insulation with values set to nil', function()
+  _G.some_global = true
+
```
Tieske, line 39: can we add a comment here that the tests in this block need to run in order (which generally is bad test design), because they test intermediate test state between tests (and hence there is no alternative).

Tieske (approved):
other than the comment; LGTM!

spec/insulate-expose_spec.lua
Henkoglobin: Ah, good point! Yepp, I'll add that.

alerque (approved):

Tieske:
thx for fixing #666 👿

Issue #668, opened 2021-09-20
New release, 2.0.1, 2.1.0?


bungle:
Several things have been added to master. Could we get a new release of busted?

Frenzie:
A new release with this line changed to `https://` would be most welcome.
https://github.com/Olivine-Labs/busted/blob/75b2c8930ed22d99dc5f6e41547ec333feee6197/busted-scm-2.rockspec#L4

```
busted scm-2 depends on mediator_lua >= 1.1.1 (not installed)
Installing https://luarocks.org/mediator_lua-1.1.2-0.rockspec
mediator_lua 1.1.2-0 is now installed in /usr/local (license: MIT <http://opensource.org/licenses/MIT>)

Cloning into 'busted'...
fatal: remote error: 
  The unauthenticated git protocol on port 9418 is no longer supported.
Please see https://github.blog/2021-09-01-improving-git-protocol-security-github/ for more information.

Error: Failed cloning git repository.
Error: Process completed with exit code 1.
```

Tieske:
@DorianGray @ajacksified The last release was almost 4 years ago. Seems like you have moved on and/or changed priorities.

Would it be possible to transfer the repositories for [busted](https://github.com/Olivine-Labs/busted), [luassert](https://github.com/Olivine-Labs/luassert), and [say](https://github.com/Olivine-Labs/say), to the [lunarmodules](https://github.com/lunarmodules) organization? This would allow us to set up a new team and keep the project going forward.

alerque:
I second that suggestion! Busted and friends are *staples* in the Lua ecosystem and deserve good active maintenance. I'm willing to pitch in some here and there as I have for other projects hosted there (Luacheck, LDoc, Penlight, etc.).

ajacksified:
@Tieske I've checked in with @DorianGray and we're happy to transfer the repositories.

ajacksified:
Done - I'll just need to set up a redirect for the docs.

ajacksified:
Also, let me know if there's anything I need to do to get you access for luarocks or anything else.

alerque:
@ajacksified The publishing to LuaRocks could me handled either of two ways:

1. You or whoever has access to the current [olivine-labs](https://luarocks.org/modules/olivine-labs) user could add an API key to that account and add it as a secret to these 3 repos.

[...]

PR #677, opened 2022-03-20
Change rockspec URL to git+https

Frenzie:
Resolves this error:

```
Cloning into 'busted'...
fatal: remote error:
  The unauthenticated git protocol on port 9418 is no longer supported.
Please see https://github.blog/2021-09-01-improving-git-protocol-security-github/ for more information.
```

Cf. <https://github.com/Olivine-Labs/busted/issues/668#issuecomment-1073215957>.

### DIFF
--- a/.github/workflows/busted.yml
+++ b/.github/workflows/busted.yml
@@ -45,18 +45,18 @@ jobs:
           luarocks install copas
           luarocks install lua-ev
           luarocks install luacov
-          luarocks install --deps-only busted-scm-2.rockspec
+          luarocks install --deps-only busted-scm-1.rockspec
 
       - name: Build ‘busted’ (bootstrap)
         run: |
           luarocks make
 
       - name: Run ‘busted’ (dogfood)
-        # disable the paths to force use of installed rockspec
-        run: busted -c -v --lpath="" --cpath="" --Xoutput "--color"
+        # disable project-local path prefixes to force use of system installation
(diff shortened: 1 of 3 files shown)


Tieske:
unfortunately all existing rockspecs must also be updated, and re-uploaded

alerque:
Yes, this is unfortunate but true.

Just a note on working around this for anybody getting stuck: the latest LuaRocks release (3.8.0) has a workaround built in to automatically update URLs in any rockspecs it processes. This won't help older LuaRocks versions trying to install order versions of rocks, but if you have the option to update LuaRocks that's one way forward.

Frenzie:
> unfortunately all existing rockspecs must also be updated, and re-uploaded

Is this a request for some kind of change in this PR? As it is nothing works and presumably all of us would like master to Just Work™. ;-)

Tieske:
@Frenzie A newer version of LuaRocks will automatically cover this case. So upgrading should get you unblocked.

@ajacksified The API key I have currently doesn't work, so I cannot fix this issue. If you can send me a new working one, I'll take care of this.

alerque (approved):
This PR has been taken care of because the SCM rockspec already reflects this change. As noted the backlog of old rockspecs will probably not be updated, although they can be installed fine winh newer Luarocks.

Thanks for the contribution.

Frenzie:
The latest rockspec plus old luarocks matters for convenience, the old rockspec not so much. :-)

PR #689, opened 2022-08-16
Update CI and and related documentation

alerque:


alerque:
I'm going to (at least temporarily) throw in the towel on setting up a GH Action for this. In particular that shouldn't hold up the next release. I have gotten it to work several ways that work for various situations, but each way has major cons. The containerized way I used for Luarocks does not work well here because Busted doesn't just need it's *own* dependencies but all the dependencies of every project it has to run on. That means dependency injection into the container and that's a mess.

I'll break out the related work out of this PR and we can look into other implementation methods that might be more generally useful to end users.

.github/workflows/busted.yml
```diff
@@ -1,46 +1,62 @@
+# ATTENTION
+#
+# This is *not* a typical busted workflow! Do not copy this to other projects!
+# Instead of installing busted from somewhere and using it on a code base,
+# we're actually building and running the current HEAD version on ourselves:
+# basically this is a combination bootstrap test and dogfooding. The upshot is
+# that we check both that everything works (and end-to-end test) and that the
+# current codebase checks against itself. This ensures we can fix a bug or make
+# a breaking change in busted without being blocked by our own CI using a
+# different busted verision.
```
Tieske, line 10:
```suggestion
# different busted version.
```

.github/workflows/busted.yml
```diff
@@ -1,46 +1,62 @@
...
 jobs:
 
   busted:
     strategy:
       fail-fast: false
       matrix:
-        luaVersion: ["5.4", "5.3", "5.2", "5.1", "luajit", "luajit-openresty"]
-        penlightVersion: ["1.9.2", "1.8.0", "1.6.0", "1.3.2"]
-    runs-on: ubuntu-20.04
+        luaVersion: [ "5.4", "5.3", "5.2", "5.1", "luajit" ] #, "luajit-openresty" ]
```
Tieske, line 24: any reason to comment out the openresty version?

.github/workflows/busted.yml
```diff
@@ -1,46 +1,62 @@
...
+      - name: Setup dependencies
+        run: |
+          sudo apt-get install libev-dev
+          luarocks install penlight ${{ matrix.penlightVersion }}
+          luarocks install luasec
+          luarocks install moonscript
+          luarocks install copas
+          luarocks install lua-ev
+          luarocks install luacov
+          luarocks install --deps-only busted-scm-2.rockspec
```
Tieske, line 48: I think the lib-ev, copas, luasec, lua-ev dependencies are based on Busted 1.x which had support for async testing. I think they can go (not 100% sure)

.github/workflows/busted.yml
```diff
@@ -1,46 +1,62 @@
...
+
+      - name: Build ‘busted’ (bootstrap)
+        run: |
+          luarocks make
+
+      - name: Run ‘busted’ (dogfood)
+        run: busted -c -v
```
Tieske, line 55:
```suggestion
        # disable the paths to force use of installed rockspec
        run: busted -c -v --lpath="" --cpath="" --Xoutput "--color"'
```

Tieske: the coverage option in `.busted` should be removed

busted-scm-2.rockspec
```diff
@@ -30,6 +38,14 @@ dependencies = {
   'mediator_lua >= 1.1.1',
 }
 
+test_dependencies = {
+  "busted",
```
Tieske, line 42: curently also luacov, unless it is removed from `.busted`

Tieske (commented):
Looks good, some small comments.

.github/workflows/busted.yml
alerque: Yes. Something is wrong with the @leafo's openresty image. It regularly but seemingly randomly crashes (maybe once every ten runs on small projects up to almost 100% on some specific large ones). There is some segfault problem that doesn't happen with openresty locally only in the GH runner images. Many other projects have temporarily disabled this too. We had to do so recently for luasocket because it was crashing there and blocking merges. Disabling it here is preemptive but I think also reasonable given the number of false positive test failures it generates elsewhere.

Tieske: ok, that makes sense then

.github/workflows/busted.yml
alerque: This project doesn't even have a `.busted`. Per explanation in [this comment](https://github.com/lunarmodules/say/pull/16#discussion_r952379886) I'm not in favor of adding one either but of course if it did the coverage option should not be enabled.

busted-scm-2.rockspec
alerque: c.f. https://github.com/lunarmodules/busted/pull/689#discussion_r952411360

.github/workflows/busted.yml
alerque: Hmm, okay I'll look into that.

Tieske: can do this later

Tieske (approved):

Tieske:
The merge after approval now leaves the CI in a failed state. Some build jobs should probably not run? @alerque

alerque:
I'll look into it.

PR #692, opened 2022-08-18
Setup repo for use *as* GH Action

alerque:
~~This is a false start, see comments in #689. A containerized Docker image is probably not the way to handle this because it needs to *execute* not just inspect the code it is testing. Probably an action that installs stuff in the host container is better.~~

A simple containerized GH Action release, per documentation best used for projects with no additional dependencies but adaptable to other uses as well.


Tieske (approved):
lgtm

PR #693, opened 2022-08-21
Update official documentation link in the README so it doesn't lead to a 404 page

demonnic:
Change does what it says on the tin. I was about to file an issue to clarify if this was the proper link when I saw a previous issue about it had already been closed and seemed to indicate this was the right one.

Rather than ask about it again I thought I'd just make a PR so future people don't get confused.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,7 +10,7 @@ busted
 busted is a unit testing framework with a focus on being **easy to
 use**. Supports Lua >= 5.1, luajit >= 2.0.0, and moonscript.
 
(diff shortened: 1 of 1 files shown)


alerque (approved):
Duplicated by prior #689, but no sweat. We still appreciate the contribution.

I wasn't the one doing it but I would guess the former "closed" issue was probably closed when somebody fixed the GitHub homepage field that shows up in the sidebar and elsewhere. They probably didn't notice the additional link in the README.

PR #695, opened 2022-08-25
chore(deps): Drop references to obsolete external dependency

### LINKED ISSUE #694: Look into current CI runner dependencies
The dependencies used in CI are more extensive than those called out in the rockspec. Look into whether this should still be the case.

See [this comment for specifics](https://github.com/lunarmodules/busted/pull/689#discussion_r952287001).

alerque:
Closes #694.

If legit. I have some doubts. The status of async support isn't perfectly clear. We still have a spec file testing various aspects of it. I assume that's why I added the copas installation in 790a9f3 some time after it was removed in e8d51ac.

### DIFF
--- a/.github/workflows/busted.yml
+++ b/.github/workflows/busted.yml
@@ -38,12 +38,8 @@ jobs:
 
       - name: Setup dependencies
         run: |
-          sudo apt-get install libev-dev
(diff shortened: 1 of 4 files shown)


alerque:
c.f. #235 for discussion on async and the removal of these deps.

PR #698, opened 2022-08-29
chore(ci) change to utfTerminal

Tieske:
to colorize output

### DIFF
--- a/.github/workflows/busted.yml
+++ b/.github/workflows/busted.yml
@@ -53,7 +53,7 @@ jobs:
 
       - name: Run ‘busted’ (dogfood)
         # disable project-local path prefixes to force use of system installation
-        run: busted --coverage --lpath="" --cpath="" -Xoutput --color
+        run: busted --coverage --lpath="" --cpath="" --output=gtest -Xoutput --color
 
       - name: Report test coverage
         if: ${{ success() && github.repository == 'lunarmodules/busted' }}


alerque:
Why did the -Xoutput=color not get applied as advertized? Should this change also be made in the Docker image so people using the GH Action version get this output by default?

https://github.com/lunarmodules/busted/blob/master/Dockerfile#L31

I'm not a huge fan of the compressed dot view for CI. It's great for local use but not my favorite is CI. The gtest with color seems like the best default for remote runners.

Tieske:
> Why did the -Xoutput=color not get applied as advertized?

Because the default outputHandler is "plainTerminal" for non-tty and "utfTerminal" for tty. The busted core hence selects "plainTerminal", which does not support the `--Xoutput` options (since it's "plain").

> The gtest with color seems like the best default for remote runners.

k, I'll update to `gtest` then

Tieske:
we could also move the options to the `.busted` file, with 2 profiles, one for local (basically `default` with an empty config) and a new profile `ci` which has all of these options. Then the only option on the Busted CLI in the CI environment would be `busted --run=ci`.

alerque:
That makes sense. If I understand what you are suggesting that just gives us two profiles for testing *this* project, it doesn't help *other* projects using the GH Action though correct? So we'd still want to configure some default arguments, probably in the Dockerfile invocation where I linked to but we could potentially setup arguments to the action that would be easier to configure than overriding the entry point.

alerque (approved):
Nevermind about messing with Docker in this PR, it seems to have color anyway but possibly for the wrong reasons. I'll look into it separately.

PR #704, opened 2022-10-10
Add manpage

linuxmaniac:
From: Nilesh Patra <npatra974@gmail.com>

https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=978760


Tieske:
since the page mentions it is generated, should we add the generation logic instead of the file itself?

alerque:
> should we add the generation logic instead of the file itself?

Yes. Lets add the code to regenerate the man page to the repo and make actually generating it part of the (automated) release process.

alerque (changes requested):
As noted before in comments but not reflected in GH's review system, this is a great contribution we'd love to have but if possible we'd actually like to make the *generation* command part of the project, not the final artifact.

PR #705, opened 2022-10-15
add Indonesian language support

amsitlab:


alerque (approved):
Thanks for taking the time to contribute! This looks good to me, but before we merge it I'd like to see matching Indonesian support added to `luassert` as well. Busted uses it for a lot of things and properly localized output will depend on having that as well. You can find [the existing localizations here](https://github.com/lunarmodules/luassert/tree/master/src/languages). Would you be able to work up a matching PR for that too?

amsitlab:
yes lunarmodules/luassert#190

Tieske:
@amsitlab thx!

PR #709, opened 2022-12-08
build: add luarocks.lock

chenrui333:
Lockfile is a way to ensure [the repeatable build](https://github.com/luarocks/luarocks/wiki/Pinning-versions-with-a-lock-file), thus adding the lockfile and update make builds to use `--pin`

also relates to https://github.com/Homebrew/homebrew-core/pull/117679

### DIFF
--- a/.github/workflows/busted.yml
+++ b/.github/workflows/busted.yml
@@ -49,7 +49,7 @@ jobs:
 
       - name: Build ‘busted’ (bootstrap)
         run: |
-          luarocks make
+          luarocks make --pin
 
       - name: Run ‘busted’ (dogfood)
         # disable project-local path prefixes to force use of system installation
--- a/.github/workflows/busted.yml
+++ b/.github/workflows/busted.yml
@@ -49,7 +49,7 @@ jobs:
 
       - name: Build ‘busted’ (bootstrap)
         run: |
-          luarocks make --pin
+          luarocks make
 
       - name: Run ‘busted’ (dogfood)
         # disable project-local path prefixes to force use of system installation
--- a/luarocks.lock
+++ b/luarocks.lock
@@ -1,0 +1,14 @@
+return {
+   dependencies = {
+      dkjson = "2.6-1",
+      lua = "5.4-1",
+      ["lua-term"] = "0.7-1",
+      lua_cliargs = "3.0-2",
+      luafilesystem = "1.8.0-1",
+      luassert = "1.9.0-1",
+      luasystem = "0.2.1-0",
+      mediator_lua = "1.1.2-0",
+      penlight = "1.13.1-1",
+      say = "1.4.1-3"
+   },
+}
--- a/try
+++ b/try
@@ -1,4 +1,4 @@
 #!/bin/sh
 luarocks remove busted --force
-luarocks make
+luarocks make --pin
 busted $@
--- a/try
+++ b/try
@@ -1,4 +1,4 @@
 #!/bin/sh
 luarocks remove busted --force
-luarocks make --pin
+luarocks make
 busted $@
--- a/README.md
+++ b/README.md
@@ -87,7 +87,7 @@ Then to reinstall and run tests:
 
 ```
 luarocks remove busted --force
-luarocks make
+luarocks make --pin
 busted spec
 ```
 
--- a/try.bat
+++ b/try.bat
@@ -1,4 +1,4 @@
 call luarocks remove busted --force
-call luarocks make
+call luarocks make --pin
 cls
 call busted %*
--- a/try.bat
+++ b/try.bat
@@ -1,4 +1,4 @@
 call luarocks remove busted --force
-call luarocks make --pin
+call luarocks make
 cls
 call busted %*


Tieske:
not sure about this. Not familiar with the lock files. If versions go out of sync, shouldn't the depedencies in the rockspec file be updated?

chenrui333:
> Not familiar with the lock files. 

lockfile is just a way to ensure repeatable builds

> If versions go out of sync, shouldn't the depedencies in the rockspec file be updated?

spec is spec (specifying version constraint), lockfile is lockfile (specifying exact version that is being used)

alerque (approved):

.github/workflows/busted.yml
```diff
@@ -49,7 +49,7 @@ jobs:
 
       - name: Build ‘busted’ (bootstrap)
         run: |
-          luarocks make
+          luarocks make --pin
```
alerque, line 52:
```suggestion
          luarocks make
```

Including a lock file is fine and many downstream situations should seriously consider using it. That being said it isn't right for every use case. This CI runner for one won't like this because we're testing a matrix of different Lua versions and dependency versions, so pinning will break all but one combination of the matrix.

luarocks.lock
```diff
@@ -0,0 +1,14 @@
+return {
+   dependencies = {
+      dkjson = "2.6-1",
+      lua = "5.4-1",
```
alerque, line 4: I understand most of the pinning, but one reason I've held off including a lock file in more projects is this Lua interpreter version pinning. Many/most projects I work on are explicitly designed and tested to run on multiple versions of Lua.

Is it possible to make `luarocks` cooperate with pinning everything except Lua itself? Does using the lock file still work if this line is dropped or is it all-or-nothing?

try
```diff
@@ -1,4 +1,4 @@
 #!/bin/sh
 luarocks remove busted --force
-luarocks make
+luarocks make --pin
```
alerque, line 3: Depending on whether the Lua interpretter can be unpinned, using pinning in these two utility scripts probably doesn't make sense.

alerque (changes requested):

.github/workflows/busted.yml
chenrui333: that makes sense, but maybe you should have at least one flow to check the lockfile though

try
chenrui333: While I have reverted the `--pin` for utility scripts, but I still strongly think they should have it. It would help with the bug reporting as well.

luarocks.lock
chenrui333: > Many/most projects I work on are explicitly designed and tested to run on multiple versions of Lua.

This is something new to me, but I already saw it yesterday. Also when did check the lua release cycle on the site, I found almost all the old versions are kind of unsupported (the last release was long time ago), I am bit confused about lua's release lifecycle.

> Is it possible to make `luarocks` cooperate with pinning everything except Lua itself? Does using the lock file still work if this line is dropped or is it all-or-nothing?

I am bit surprised that `luarocks` lock `lua` version though, it can be outside as engine constraint (like npm, or other build tools). But for what I see in the other build tools, it is all or nothing.

I am gonna search a bit and see if anyone feedback on this `lua` line.

alerque: Lua is a pretty fragmented ecosyntem with no single canonical direction. The main issue is that even though development continues in one direction (call this where Lua 5.4 is), it also got forked back in 5.1 to create LuaJIT. Since the JIT fork is substantially faster it has kept up a life of it's own with new features that diverged even from 5.1. Even that JIT project has been forked with another popular project (OpenResty) bringing another mix of features and tradeoffs. Aditionally Lua runtimes are embedded in lots of other software (e.g. Pandoc), ofter with version specific bindings. Complicating matters is the widespread use of distros with LTS releases that have old versions out there. Hence many projects in the Lua realm have to either pick a lane or deliberately use a lowest-common-denominator approach to run on as many as possible.

Busted itself *must* take the latter approach, especially since it gets used to test projects that are limited to various different interpreters.

alerque: @hishamhm Any comment on why the VM is included in the `make --pin` output this way?

chenrui333: Gonna wait for the comment before doing anything.

Tieske:
>>    Many/most projects I work on are explicitly designed and tested to run on multiple versions of Lua.
>
>This is something new to me, but I already saw it yesterday. Also when did check the lua release cycle on the site, I found almost all the old versions are kind of unsupported (the last release was long time ago), I am bit confused about lua's release lifecycle.

The main reason for this is that Lua is an embedded language. (not as in hardware, but as embedding it in other applications to make them scriptable). This typically means that those host applications have their own ecosystem with Lua code. And hence upgrading the Lua version does not make sense as it is extremely intrusive. So unless you really need a feature (UTF-8 for example), you stick to the original version you started with.

It's the embedded nature that makes this different from other languages, where the tendency is to stay current.

luarocks.lock
hishamhm: @alerque Because this is intended as a way to get more stable builds via a fixed set of dependencies — some rocks will fetch different dependencies based on Lua version, so a given dependency graph is actually dependent on the Lua version (for example, IIRC, some of lhf's modules have different versions for different Lua versions).

PR #711, opened 2023-03-17
fix(luajit): ensure that ffi loaded libs will not get GC'ed

Tieske:
This caused an issue when busted would clear `_G` and the lib was unloaded whilst userdata's were still floating around with references to the now unloaded lib. This caused segfaults.


bungle (approved):

alerque (approved):

Tieske:
I'll merge this shortly, waiting for some extra confirmation it solved the issue

Tieske:
confirmed!

PR #713, opened 2023-03-31
fix(rockspec): remove luafilesystem, not a dependency

dundargoc:
"lfs" isn't being required in the codebase.

### DIFF
--- a/busted-scm-1.rockspec
+++ b/busted-scm-1.rockspec
@@ -35,7 +35,6 @@ description = {
 dependencies = {
   'lua >= 5.1',
   'lua_cliargs = 3.0',
-  'luafilesystem >= 1.5.0',
   'luasystem >= 0.2.0',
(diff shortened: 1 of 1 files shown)


alerque:
I *think* (but it is kind of disconcerting that CI runs did not show this) that `lfs` gets checked for in one of our dependencies where it is an optional dependency—and that we need the optional dependency required. On the other hand we also depend on Penlight, which in turn requires LuaFileSystem, so perhaps that's why CI didn't surface a problem. It ends up on the system anyway even if we do drop this here.

I'm happy to drop it if that is not the case, but it may not be quite as simple as grepping for *lfs*.

dundargoc:
I've grepped for `lfs` not only in busted but in every package it includes transitively. It is only used in `penlight`.

alerque (approved):

Tieske (approved):

PR #714, opened 2023-04-07
chore(readme): fix the introductory example

C3pa:
There is no such table as "spy_on" inside "spy".

image


alerque (approved):

alerque (approved):
Thanks for taking the time to contribute!

Tieske (approved):

PR #720, opened 2023-05-12
fix(testing): set LUA_PATH in try script

hanshuebner:
This change makes the `try` script work in environments that don't already have `LUA_PATH` set up the way `busted` needs.  The `try` script should eventually be replaced by a `Makefile` that also verifies external dependencies, but this will help for now.

### DIFF
--- a/try
+++ b/try
@@ -1,4 +1,4 @@
 #!/bin/sh
 luarocks remove busted --force
 luarocks make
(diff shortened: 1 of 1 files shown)


Tieske:
Thx! We'd indeed be better off implementing a proper Makefile

Issue #729, opened 2023-09-16
Fails to encode results to json due to non-string error objects being raised


svermeulen:
I have a custom error table that I pass to `error(foo)`.  When I do this in a busted test, and then pass `--output=json`, it will just print nothing, due to an error that occurs in json.encode where it is unable to serialize function objects.  I'm guessing that busted should run `tostring(error_obj)` somewhere to address this problem

YaroSpace:
Reopening the issue:
I am getting `Failed to encode test results to json: type 'function' is not supported by JSON` when running a suite with pending tests.  Pending tests have an attribute  - `default_fn` which is a function (defined as `function() end` in busted `init.lua:71', that cannot be encoded into json.

alerque:
PRs welcome. Just papering over this with `tostring()` is probably not going to help though because comparing the string ID for two functions (even with identical content) is going to turn up `!=`. Some extra consideration might be needed.

YaroSpace:
Ok. Will investigate further)

YaroSpace:
> I have a custom error table that I pass to `error(foo)`. When I do this in a busted test, and then pass `--output=json`, it will just print nothing, due to an error that occurs in json.encode where it is unable to serialize function objects. I'm guessing that busted should run `tostring(error_obj)` somewhere to address this problem

@svermeulen I know it has been a while, but can you tell me how to reproduce your issue?  I am trying to pass a custom error object to `error()`, but I am not getting any problems.

PR #730, opened 2023-09-18
fix(json-output): convert error object to string before encoding

svermeulen:
Fixes issue #729

Also worth noting that a better way to handle json encode errors would probably be to report this via the json object instead of just outputting the error directly, but at least this way there is some output indicating the problem (previously no output about error would occur)

### DIFF
--- a/busted/outputHandlers/json.lua
+++ b/busted/outputHandlers/json.lua
@@ -14,9 +14,7 @@ return function(options)
       errors = handler.errors,
       duration = handler.getDuration()
     }
-    local ok, result = pcall(function()
-      return json.encode(error_info)
-    end)
+    local ok, result = pcall(json.encode, error_info)
 
     if ok then
       io_write(result)
--- a/busted/outputHandlers/json.lua
+++ b/busted/outputHandlers/json.lua
@@ -7,13 +7,23 @@ return function(options)
   local handler = require 'busted.outputHandlers.base'()
 
   handler.suiteEnd = function()
-    io_write(json.encode({
+    local error_info = {
       pendings = handler.pendings,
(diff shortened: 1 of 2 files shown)


busted/outputHandlers/json.lua
```diff
@@ -7,13 +7,23 @@ return function(options)
...
       pendings = handler.pendings,
       successes = handler.successes,
       failures = handler.failures,
       errors = handler.errors,
       duration = handler.getDuration()
-    }))
+    }
+    local ok, result = pcall(function()
+      return json.encode(error_info)
+    end)
```
alerque, line 19: The `pcall()` function will pass on extra arguments to the function referenced in the first argument, so you don't need an anonymous closure for this situation:

```suggestion
    local ok, result = pcall(json.encode, error_info)
```

busted/outputHandlers/json.lua
```diff
@@ -7,13 +7,23 @@ return function(options)
...
+
+    if ok then
+      io_write(result)
+    else
+      io_write("Failed to encode test results to json: " .. result)
```
alerque, line 24: Question out of ignorance here, but is `result` as used here already guaranteed to be a string, or should it be cast `tostring()` just in case?

alerque (changes requested):
Thanks for taking the time to contribute. This looks pretty good in general, just a small nit and a question...

busted/outputHandlers/json.lua
svermeulen: Ah yeah true.  Fixed

alerque (approved):

Tieske (approved):

PR #732, opened 2023-11-23
Hack around cliargs to provide "help" flag

### LINKED ISSUE #697: Output --help to STDOUT
Currently the output of `--help` is directed to STDERR which is wrong per GNU guidelines and quite annoying when trying to look for something. One has to redirect `2>&1` before greping or paging the output.

TheMagician23:
Fixes #697

This change adds a check on the error returned from the `cliargs` library, and provides a "help" argument to the consumer code in `busted/runner.lua`.
It is based on the workaround used with the same library here: https://github.com/sile-typesetter/sile/pull/1737/files
except that the code in `busted/runner.lua` probably won't need to change if `cliargs` ever supports adding a `--help` flag in the "correct" way.

### DIFF
--- a/busted/runner.lua
+++ b/busted/runner.lua
@@ -41,6 +41,11 @@ return function(options)
     exit(1, forceExit)
   end
 
+  if cliArgs.help then
+    io.stdout:write(cliArgs.helpText)
+    exit(0, forceExit)
+  end
+
   if cliArgs.version then
     -- Return early if asked for the version
     print(busted.version)
--- a/busted/runner.lua
+++ b/busted/runner.lua
@@ -42,13 +42,13 @@ return function(options)
   end
 
   if cliArgs.help then
-    io.stdout:write(cliArgs.helpText)
+    io.stdout:write(cliArgs.helpText .. '\n')
     exit(0, forceExit)
   end
 
   if cliArgs.version then
     -- Return early if asked for the version
-    print(busted.version)
+    io.stdout:write(busted.version .. '\n')
     exit(0, forceExit)
   end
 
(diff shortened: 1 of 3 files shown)


alerque:
The upstream for cli_args has (since this PR was started) migrated the project to this org. I proposed this a long time ago but this PR made me ping the author again and this time it happened. Now that we *can fix upstream* I think we *should* fix this issue at the source and then just *use* it correctly in this project.

TheMagician23:
Hi Caleb.

This is fantastic news! I saw that lua_cliargs badly needed to be migrated to lunarmodules when I did my intial research, I'm glad you managed to make it happen.
I'll write a fix for the issue upstream soon; once it's done I'll comment on the issue in the busted repo with a link to the upstream fix.
Once it's in, you can throw away my original PR. Let me know if you want me to write a proper usage patch for busted after that fix goes in.

- The Magician

TheMagician23:
Okay, I've written a PR to improve the help system in lua_cliargs: https://github.com/lunarmodules/lua_cliargs/pull/78

alerque (approved):

busted/runner.lua
```diff
@@ -41,6 +41,11 @@ return function(options)
     exit(1, forceExit)
   end
 
+  if cliArgs.help then
+    io.stdout:write(cliArgs.helpText)
```
Tieske, line 45: just checking: does the help text include a trailing newline? `io.stdout:write()` will not automatically append one like `print()` does.

alerque: Good catch, it does *not* include a trailing newline and is used elsewhere inline so it doesn't make sense to add it to the content. I just addressed this by explicitly concating a newline to it as we write it out, and also normalized away the one use of `print()` instead of `io.<streat>:write()` like we are using everywhere else in this module.

Tieske (approved):

PR #733, opened 2023-12-12
Unpin cliargs dependency

alerque:
I haven't looked into *why* this was pinned in the first place yet, but whatever the reason I think we need to fix it at the source, not this way. This is making it hard to even test the cliargs module, much less release it.

### DIFF
--- a/busted-scm-1.rockspec
+++ b/busted-scm-1.rockspec
@@ -34,7 +34,7 @@ description = {
 
 dependencies = {
   'lua >= 5.1',
-  'lua_cliargs = 3.0',
+  'lua_cliargs >= 3.0',
   'luasystem >= 0.2.0',
   'dkjson >= 2.1.0',
   'say >= 1.4-1',


alerque:
This hard pinning of versions goes back all the way to 02f26cd and has no apparent rational there either.

busted-scm-1.rockspec
```diff
@@ -34,7 +34,7 @@ description = {
 
 dependencies = {
   'lua >= 5.1',
-  'lua_cliargs = 3.0',
+  'lua_cliargs >= 3.0',
```
Tieske, line 37:
```suggestion
  'lua_cliargs ~> 3',
```
Wouldn't it be safer to pin to the major version?

alerque: Not "safer" for sure but maybe "safe enough". If we follow semver specs breaking changes should only be in major versions. I think our big question is whether we're going to consider the change in clirags to `--help` handling so it doesn't throw an error exit status as a breaking change or not.

Tieske: I'd vote no...

alerque: That was my first thought too, but we have a v4 coming up pretty soon to fix the SPLAT defaults, so if we *wanted* to hold off until it would be in a breaking release anyway we could. Personally I think the behavior is so *wrong* fixing it as a bug should be the right thing to do even if it does break a few apps expectations downstream.

Tieske: @alerque your call, I'm fine either way.

rautyrauty:
@alerque merge pls

SlySven:
This is now a breaking issue as `lua_cliargs` has had a patch level revision:
```bash
stephen@Ripley:~/src/mudlet$ luarocks list --local  --outdated

Outdated rocks:
---------------

lua_cliargs
   3.0-2 < 3.0.2-1 at https://luarocks.org

stephen@Ripley:~/src/mudlet$ luarocks install --local lua_cliargs
Installing https://luarocks.org/lua_cliargs-3.0.2-1.src.rock

lua_cliargs 3.0.2-1 depends on lua >= 5.1 (5.1-1 provided by VM)
lua_cliargs 3.0.2-1 is now installed in /home/stephen/.luarocks (license: MIT)

Checking stability of dependencies in the absence of
lua_cliargs 3.0-2...

Will not remove lua_cliargs 3.0-2.
Removing it would break dependencies for: 
busted 2.2.0-1

Use --force to force removal (warning: this may break modules).

stephen@Ripley:~/src/mudlet$
```

hishamhm:
> This is now a breaking issue as lua_cliargs has had a patch level revision:

This was not an error message; it kept both versions `3.0-2` and `3.0.2-1` around in your system because of the pinned dependency.

> Use --force to force removal (warning: this may break modules).

The warning was relative using to `--force` to remove the old one, in case that wasn't clear.

@alerque I have no idea why that was pinned. Should be safe to unpin.

hishamhm (approved):

PR #734, opened 2024-01-15
opt: junit support split out; optimize display format

yeshan333:
**Features**:

- Support outputting Junit XML reports in suite dimension. When the scale of test cases is large, combining with some CI systems (eg: [Jenkins CI JUnit Realtime Test Reporter](https://plugins.jenkins.io/junit-realtime-test-reporter/)) can enable real-time viewing of test reports.

**Optimizations**:
- The current implementation is not friendly to the display of CI systems, and the hierarchy is not clear enough.
  - Demo Link: https://jenkins.shan333.cn/job/busted_junit_demo/2/testReport/busted_demo.spec/ 
  - before
- Reason: CI System use the last one . to distinguish between package name and class name.

After optimizations:

- Demo Link: https://jenkins.shan333.cn/job/busted_junit_demo/3/testReport/spec/
- opts

Reproduction demo: https://github.com/yeshan333/busted_junit_demo

### DIFF
--- a/busted/outputHandlers/junit.lua
+++ b/busted/outputHandlers/junit.lua
@@ -23,11 +23,16 @@ return function(options)
     })
   }
   local output_file_name
+  local enable_split_output_xml
   local stack = {}
   local testcase_node
   if 'table' == type(options.arguments) then
-    --the first argument should be the name of the xml file.
-    output_file_name = options.arguments[1]
+    if options.arguments[1] == "true" then
+      enable_split_output_xml = true
+    else
+      -- will output to the sepcific xml file
+      output_file_name = options.arguments[1]
+    end
   end
 
   handler.suiteStart = function(suite, count, total)
@@ -67,6 +72,27 @@ return function(options)
     top.xml_doc.attr.failures = top.xml_doc.attr.failures + suite_xml.xml_doc.attr.failures
     top.xml_doc.attr.skip = top.xml_doc.attr.skip + suite_xml.xml_doc.attr.skip
 
+    top.xml_doc.attr.time = elapsed(top.start_tick)
+
+    if enable_split_output_xml ~= nil then
+      local output_string = xml.tostring(top.xml_doc, '', '\t', nil, false)
+      local test_suit_file = suite['file']
+
+      local write_file
+      if test_suit_file ~= nil then
+        write_file = string.gsub(test_suit_file[1].name, "%.[^.]+$", ".xml")
+      else
+        write_file = "no_match_cases.xml"
+      end
+
+      local file = io_open(write_file, 'w+b' )
+      if file then
+        file:write(output_string)
+        file:write('\n')
+        file:close()
+      end
+    end
+
     return nil, true
   end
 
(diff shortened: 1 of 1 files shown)


alerque:
Are there any junit consumers that would *not* benefit from this? As in, is there an argument for this being the default?

yeshan333:
> Are there any junit consumers that would _not_ benefit from this? As in, is there an argument for this being the default?
@alerque 

This is very useful for users who use web UI to view test reports in CI System（eg：Jenkins. Junit is one of the most widely used testing frameworks in the Java ecosystem, while Jenkins is the most widely used automation system in the Java ecosystem.）, But this will not have any impact on users who enjoy using terminals to view results.

**Java's classes use （.） De-district the hierarchical structure of the project.**

The original implementation resulted in an ugly use case hierarchy for CI splitting.

the `element.trace.short_src` are usually forms such as `case_spec.lua`. This means that `lua` has been recognized as a class in the Java ecosystem.

https://github.com/lunarmodules/busted/blob/02f31a9c103a44e166617cfdb6ba1b8994a9c912/busted/outputHandlers/junit.lua#L106

For outputting test cases in the `suiteEnd` stage, when the scale of test cases is very large, it is convenient to view some results in real time (perhaps you may ask: is it not good to directly specify the execution of a certain test case, but for projects where some test cases will affect each other, this helps to identify problem cases or test small earlier)
For outputting test cases in the xx stage, when the scale of test cases is very large, it is considered to view some results in real time (per hats you may ask, it is not good to directly specify the execution of a certain test case, but for projects where some test cases will affect each other, this helps to identify problem cases or find [test smells](http://xunitpatterns.com/TestSmells.html) problems earlier)

busted/outputHandlers/junit.lua
```diff
@@ -101,10 +127,37 @@ return function(options)
...
+      junit_classname = junit_report_package_name .. "." .. junit_class_name ..":" .. element.trace.currentline
+    else
+      junit_classname = junit_class_name ..":" .. element.trace.currentline
+    end
+
     testcase_node = xml.new('testcase', {
-      classname = element.trace.short_src .. ':' .. element.trace.currentline,
-      name = handler.getFullName(element),
+      -- junit report uses package names and class names to structurally display result.
+      classname = junit_classname,
```
yeshan333, line 159: Reason for modification reference:

https://github.com/jenkinsci/junit-plugin/blob/67297b3bcae8ec8422b1b9a965595f7187fd9318/src/main/java/hudson/tasks/junit/CaseResult.java#L182

alerque:
Yes I get how this improves output for web UIs setup for Java, my question was about whether the output would be *worse* anywhere else if we made it the default. Is junit ever useful on a terminal for example? Or would folks normally be using other outputters there anyway?

yeshan333:
> Yes I get how this improves output for web UIs setup for Java, my question was about whether the output would be _worse_ anywhere else if we made it the default. Is junit ever useful on a terminal for example? Or would folks normally be using other outputters there anyway?

The impact is minimal. The junit terminal output works well. I have some tests on Github Actions and Gitlab CI, you can take a look:
- GitHub Actions:
  - terminal output: https://github.com/yeshan333/busted_junit_demo/actions/runs/7539484235/job/20522081046
  - get report: https://github.com/yeshan333/busted_junit_demo/actions/runs/7539484235/job/20522085626
- GitLab CI Demo：https://gitlab.com/yeshan333/busted_junit_demo/-/pipelines/1139567228/test_report?job_name=golang

alerque (approved):

PR #740, opened 2024-03-09
Implement loader for Fennel

### LINKED ISSUE #738: Fennel loader?
Hello,

I would like to write tests in [Fennel](https://fennel-lang.org/). My understanding is that this would be possible by implementing a Fennel runner. Looking at the source code it seems like the files under `busted/modules/files` implement loaders for Lua, Moonscript and Terra. A loader is a table with two functions `load` and `match`, but there is no explanation as to what interface these two functions must implement.

I would like to try my hands at writing a Fennel runner. Is there some explanation, or can someone here please elaborate on the protocol that needs to be implemented?

HiPhish:
Closes #738

Tests can be written in Fennel if `fennel` is among the loaders. Currently the traceback is wrong for failures and pending tests, but it is wrong for tests written in Moonscript as well, so this seems to be a general problem with busted.

It would have been cool if we could make `it`, `describe` and so on into macros to not have to manually wrap the test body inside a thunk every time.

```fennel
;; With macro
(it "Adds two numbers"
  (assert.are.equal 5 (+ 2 3)))

;; Without macro
(it "Adds two numbers"
  (fn []
    (assert.are.equal 5 (+ 2 3))))
```

This macro can accomplish it when added to the beginning of a test:

```fennel
(macro it [description & body]
  `(let [fenv# (or (and getfenv (getfenv)) _ENV)]
     ((. fenv# :it) ,description
                (fn [] ,(unpack body)))))
```

However, the source map produces rather weird mappings from Lua to Fennel (it maps the assertion to the body of the macro definition), so for now I won't pursue this.

### DIFF
--- a/busted-scm-1.rockspec
+++ b/busted-scm-1.rockspec
@@ -82,6 +82,7 @@ build = {
     ['busted.modules.files.lua']              = 'busted/modules/files/lua.lua',
     ['busted.modules.files.moonscript']       = 'busted/modules/files/moonscript.lua',
     ['busted.modules.files.terra']            = 'busted/modules/files/terra.lua',
+    ['busted.modules.files.fennel']           = 'busted/modules/files/fennel.lua',
 
     ['busted.outputHandlers.base']            = 'busted/outputHandlers/base.lua',
     ['busted.outputHandlers.utfTerminal']     = 'busted/outputHandlers/utfTerminal.lua',
--- a/busted/modules/files/fennel.lua
+++ b/busted/modules/files/fennel.lua
@@ -1,0 +1,102 @@
+-- NOTE: the traceback of `pending` and `error` is broken, and there does not
+-- seem to be anything we can do about it.  The Moonscript loader has the same
+-- problem.
+
+
+local path = require 'pl.path'
+local ok, fennel = pcall(require, 'fennel')
+
+local MESSAGE_PATTERN = '^([^\n]-):(%d+): (.*)'
+local MESSAGE_TEMPLATE = '%s:%d: %s'
+local source_maps = {}
+
+local ret = {}
+
+-- A Fennel variant of the standard Lua loadstring function; registers a source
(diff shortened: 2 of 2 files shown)


alerque (changes requested):
At first blush this looks like it's going to be fine. I don't have any Fennel expertise but it isn't like this is going to break anything *else*, and I don't expect any real reason to not include such a module if Fennel folks say it is useful.

I'll review details later this week (feel free to poke again if this falls off my radar), but to quick things:

1. I've been roughly following the [conventional commit spec](https://www.conventionalcommits.org/en/v1.0.0/), so this commit message should be `feat(core): Implement loader for Fennel` or something like that.
2. We should probably add a test. Somehow...obviously we're going to have a dependency problem with that but we can fix that in CI and maybe make it an optional test for local testing for folks that don't have Fennel setup. Is their prior art for testing our other loaders?

HiPhish:
Thank you for your response. I have amended the commit message.

> We should probably add a test.

Maybe I am missing something here, but it does not look like there are any loader tests.  There is the file `spec/modules/file_loader_spec.lua`, but all it does is require and call a module, but it does not do anything with the results.

> and maybe make it an optional test for local testing for folks that don't have Fennel setup.

That's simple enough, we can `pcall(require, 'fennel')` in the test; if the call errors then Fennel is not available and we can skip the Fennel loader.

> I don't have any Fennel expertise

[Fennel](https://fennel-lang.org/) is pretty much just Lua with Lisp notation (some special forms plus a macro system), so it maps very closely to Lua.

HiPhish:
> feel free to poke again if this falls off my radar

*poke* @alerque

offsetcyan:
poke @alerque

emmabukacek:
Poke @alerque 🐸

noncom:
poke @alerque 😄

alerque (approved):

noncom:
❤️

PR #752, opened 2025-02-26
fix(l10n): removed inappropriate line

bayram-dev:


alerque:
This is a sensitive issue, and this post isn't meant to resolve everything. Just some thoughts about why I'm going to close this.

* These strings are not meant to be translations. The way they are used is a random set of quips that each language might consider apropos of a positive outcome.
* These quips are not presented to users of other languages. They are not the same across all the languages.
* This particular set was contributed by the translator for the actual Ukrainian strings, and it seems likely they were doing something their language group and culture deemed appropriate.
* It seems a little suspect to me, and all things being equal I'd like to see these quips stay away from anything overtly political, religious, or otherwise likely to offend.
* Your GitHub account is advertised as being from a country that is at war with the country that speaks the language in question. Even if you were just suggesting a verifiable typo fix or something I would subject such a submission to extra scrutiny just to not step on toes.
* If this particular quip were to be removed or substituted in the UA data set, the change at least should come from UA speaking/affiliated users—even just to avoid an appearance of dragging external political factors into a FOSS development cycle where it doesn't *need* to be so.

PR #753, opened 2025-06-25
feat: support multiple finally calls within a single test

### LINKED ISSUE #558: Support multiple `finally` calls within a single test
Currently calling `finally` a second time overwrites the first finalizer, so that only the last `finally` call in the test has an effect. This is inconvenient when there are several separately acquired resources that need to be finalized.

For example, this prints only `Closing file2` but not `Closing file1`:

```lua
local function open_file(path, mode)
   local f = assert(io.open(path, mode))
   finally(function()
      print("Closing " .. path)
      f:close()
   end)
   return f
end

it("test", function()
   local f1 = open_file("file1", "w")
   local f2 = open_file("file2", "w")
   -- Do some checks
end)
```

goldenstein64:
Fixes #558

This should allow for adding multiple `finally` blocks to an `it` block. Blocks are invoked in the reverse order they were added. Newer blocks are typically added from a deeper abstraction, so they must make less assumptions about the ordering of `finally` blocks.


goldenstein64:
I want to note that I haven't been able to get all unrelated tests to pass.

Tieske:
@goldenstein64 can you rebase and pick up the changes in #781 ? I had copilot do a rebase, but it stole the git-attribution from you in the process.

goldenstein64:
I am busy at the moment, but I will look at it in a couple hours.

goldenstein64:
Okay, well, I rebased #781 onto this branch, but I did the same to the rest of master. It's probably fine since it makes a distinction between who authored and who committed.

Tieske (approved):

Tieske:
Thx @goldenstein64 !

PR #760, opened 2025-12-29
fix(lua5.5): add local variable

Tieske:
Lua 5.5 has the loop-variable defined as a const, and hence modifying now throws an error.

There might be more instances, but since this is a catch-22 with the Penlight dependency (which is tested using Busted again), its a whack-a-mole for starters...


Tieske:
CI failures are existing, and seem unrelated, hence merging

PR #762, opened 2026-01-04
chore(deps): bump penlight to 1.15.0

Tieske:
first Penlight version compatible with Lua 5.5

### DIFF
--- a/.github/workflows/busted.yml
+++ b/.github/workflows/busted.yml
@@ -21,8 +21,8 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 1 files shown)


Tieske:
@alerque the failures for 5.5 are the same as for older Lua versions, so no regressions (just existing failures).

With Lua 5.5 there also is a LuaCov failure, but that's a LuaCov issue. LuaCov also relies on Busted for its CI, so imho we should merge and release this PR. From there we can fix the other projects depending on Busted.

alerque (approved):

PR #763, opened 2026-01-05
Changed the way in which existence of ffi is detected, so that external ffi libraries can also be picked up

sudheerhebbale:
In the file busted/luajit.lua

The presence of ffi as a module is tested using the following code

```
local isJit = (tostring(assert):match('builtin') ~= nil)

if not isJit then
  return function() end
end
```

This works well for luaJit, which is compatible with lua5.1; luaJit, however, is incompatible with later versions lua5.2, lua5.3 and lua5.4

To address this, I have been maintaining a ffi library <https://github.com/Tekenlight/luaffifb> which is a fork from <https://github.com/facebook/luaffifb>. This is supported on Linux x86_64 and ARM64 and OSX x86_64 and ARM64 platforms.

While testing a simple test case 

With two identical files test1.lua and test2.lua with content:

```
local ffi = require "ffi"
describe("Test", function() end)
```

It is resulting in a core double free error as below

```
$ busted --lua=lua5.3 test1.lua test2.lua

0 successes / 0 failures / 0 errors / 0 pending : 0.003987 seconds
lua5.3(31915,0x208f460c0) malloc: Double free of object 0x120804b20
lua5.3(31915,0x208f460c0) malloc: *** set a breakpoint in malloc_error_break to debug
```

I have done a fix to this issue, in busted/luajit.lua module where the testing for existence of ffi is changed to look for external ffi libraries apart from luaJit only.

Can you please merge that PR and publish the fix on luarocks, if done it will help myself as well as other users of the library I have been maintaining.

### DIFF
--- a/busted/luajit.lua
+++ b/busted/luajit.lua
@@ -1,6 +1,45 @@
+local function detect_ffi()
+  local ok, ffi = pcall(require, "ffi")
+  if not ok then
+    return { ok = false, err = tostring(ffi) }
+  end
+
+  local is_luajit = type(jit) == "table" and type(jit.version) == "string"
+
+  -- If ffi comes from the filesystem, these will usually resolve.
+  local lua_file  = package.searchpath("ffi", package.path)
+  local c_file    = package.searchpath("ffi", package.cpath)
+
+  -- Another hint: which searcher provides the loader?
+  local loader_from
+  for i, s in ipairs(package.searchers or package.loaders) do
+    local f, extra = s("ffi")
+    if type(f) == "function" then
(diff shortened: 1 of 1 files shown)


Tieske:
why should the FFI be detected in this case? it is not detected, since it is not the LuaJIT ffi, which is correct.

Use a helper script to load the ffi module before isolation kicks in.

If your FFI methods need patching (because of the same globals issue LuaJIT ffi suffers from), then you can optionally patch those methods in the same helper script. For a helper example see: https://github.com/lunarmodules/busted/issues/756#issuecomment-3719682695

Issue #770, opened 2026-03-14
suppress-pending doesn't seem to work in .busted files


ladybunne:
With a `.busted` file containing:

```
return {
    _all = {
        suppress-pending = true
    },
}
```

Running `busted` results in:

```
failed loading config file `.busted`: .busted:3: '}' expected (to close '{' at line 2) near '='
```

I've tested with other options and they seem to work (like `verbose`), so I think it's just this one.

I'm using busted version 2.3.0 and Lua version 5.4.8.

Tieske:
the file is a Lua file that is being read, so the error you are seeing is the Lua parser failing to load it, because of the "-" character which isn't allowed in Lua names.

You can make it a valid Lua file like this:
```lua
return {
    _all = {
        ["suppress-pending"] = true,
    },
}
```

ladybunne:
Right! I think the docs should be updated with a line covering this, since if you just follow them as directed you'd run into this issue as I have.

Thanks for the workaround!

Tieske:
Docs are in the `gh-pages` branch, PR's are appreciated.

closing this as resolved.

PR #773, opened 2026-03-16
Update docs: add note about .busted files and options containing hyphens

ladybunne:
Hi! First time contributing. This is a PR to address #770 as suggested by @Tieske.

I've modified the `.busted` file sample code to include a line covering this case, and a paragraph below to explain the case (that it's due to Lua parsing quirks).

Happy to tweak wording as needed, hopefully this is what you're looking for!

### DIFF
--- a/index.html
+++ b/index.html
@@ -268,7 +268,8 @@ OPTIONS:
     coverage = true
   },
   default = {
-    verbose = true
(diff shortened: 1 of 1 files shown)


Tieske (approved):
lgtm! thanks for the contribution

PR #777, opened 2026-06-02
Drop lua-term since lua-system has the same functionality

### LINKED ISSUE #771: refactor; drop lua-term since lua-system has the same functionality
`lua-term` is used for the `isatty` function. Yet recent versions of LuaSystem (which busted also depends upon) have this function as well.

By switching to the luasystem implementation, we can drop the `lua-term` dependency.

IlyasYOY:
Removed the `lua-term` as described here: https://github.com/lunarmodules/busted/issues/771#issuecomment-4602083767

- Ran `just check`, no new test failures. Seems like the change in the infrastructure layer where are no tests.
- Ran `rg` to search `'test` in the code, no more entries found.
- Cloned `colors.lua` from `lua-term` to `busted/outputHandlers/colors.lua`. Mentioned it in docs. Kept it AS-IS, didn't change a thing. I might remove unused code if you think it'd be better.
- I didn't create a new rockspec, because that it's better be done before the release.

The only thing left to do is: 

```diff
--- rockspecs/busted-2.3.0-1.rockspec	2026-06-02 14:59:47
+++ rockspecs/busted-2.3.1-1.rockspec	2026-06-02 17:59:25
@@ -1,5 +1,5 @@
 local package_name = "busted"
-local package_version = "2.3.0"
+local package_version = "2.3.1"
 local rockspec_revision = "1"
 local github_account_name = "lunarmodules"
 local github_repo_name = package_name
@@ -34,11 +34,10 @@
 dependencies = {
   'lua >= 5.1',
   'lua_cliargs >= 3.0',
-  'luasystem >= 0.2.0',
+  'luasystem >= 0.3.0',
   'dkjson >= 2.1.0',
   'say >= 1.4-1',
   'luassert >= 1.9.0-1',
-  'lua-term >= 0.1',
   'penlight >= 1.15.0',
   'mediator_lua >= 1.1.1',
 }
```

Closes: https://github.com/lunarmodules/busted/issues/771

---

I'm not sure about AI policy, so I write it here: 

- commit messages are made using codex, 
- everything else I did myself. 

I'm eager to make more contributions after this one, so it'd be good to understand the AI-usage policy. I don't want to violate it.

### DIFF
--- a/busted-scm-1.rockspec
+++ b/busted-scm-1.rockspec
@@ -35,11 +35,10 @@ description = {
 dependencies = {
   'lua >= 5.1',
   'lua_cliargs >= 3.0',
-  'luasystem >= 0.2.0',
+  'luasystem >= 0.3.0',
   'dkjson >= 2.1.0',
   'say >= 1.4-1',
   'luassert >= 1.9.0-1',
-  'lua-term >= 0.1',
   'penlight >= 1.15.0',
   'mediator_lua >= 1.1.1',
 }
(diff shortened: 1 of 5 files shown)


IlyasYOY:
Updated `busted-scm-1.rockspec`, now it's up-to-date with the implementation. 

Checked the tests after reinstalling deps. 

PS. There is luacov is missing, so the `just check` fails if I don't install it manually: `luarocks --tree lua_modules install luacov`, should we add luacov in test dependencies?

